PR #12, opened 2018-02-06
ED-3333 automated update script

eadpearce:
https://uktrade.atlassian.net/browse/ED-3333


scripts/upgrade_header_footer.py
```diff
@@ -0,0 +1,122 @@
...
+        "requirements.in",
+        "requirements_test.txt",
+        "requirements_test.in",
+    ]
+
+    exp = r'(?:directory-components\.git@v)(\d*\.\d*\.\d)'
+    get_version = r'\d*\.\d*\.\d'
+
+
+class Colours:
```
richtier: [ansi-colours for Python](https://github.com/jonathaneunice/colors/) could be used.

I :heart: delivering the greatest functionality with the lowest complexity code. Using an existing third-party library reduces the amount of code we write, test, review, and maintain - so we can deliver more functionality with less effort.

PR #15, opened 2018-02-20
ED-3232 use base url for header/footer urls

eadpearce:
https://uktrade.atlassian.net/browse/ED-3232


directory_components/context_processors.py
```diff
@@ -50,12 +51,12 @@ def urls_processor(request):
...
             "title": "Home",
             "id": "home-link",
-            "url": get_url("GREAT_HOME")
+            "url": get_url("HEADER_FOOTER_URLS_GREAT_HOME")
         },
         "custom": {
             "title": "Your export journey",
             "id": "custom-page-link",
-            "url": get_url("CUSTOM_PAGE")
+            "url": "{}custom/".format(get_url("HEADER_FOOTER_URLS_GREAT_HOME"))
```
richtier: could you use `urljoin` (the url equivalent of `os.path.join`) please.

```
from urllib.parse import urljoin

urljoin(get_url("HEADER_FOOTER_URLS_GREAT_HOME"), 'custom')
```

directory_components/context_processors.py
```diff
@@ -163,31 +175,49 @@ def urls_processor(request):
...
+    header_footer_urls = {
+        "home": get_url("HEADER_FOOTER_URLS_GREAT_HOME"),
+        "fab": get_url("HEADER_FOOTER_URLS_FAB"),
+        "soo": get_url("HEADER_FOOTER_URLS_SOO"),
+        "events": get_url("HEADER_FOOTER_URLS_EVENTS"),
+        "contact_us": get_url("HEADER_FOOTER_URLS_CONTACT_US"),
+        "dit": get_url("HEADER_FOOTER_URLS_DIT"),
+    }
+    return {
+        'header_footer_urls': header_footer_urls
```
richtier: as these urls are to other services used outside of header/footer templates, perhaps name it something `directory_components_urls`, rather than `header_footer_urls` (as the current name is a bit misleading).

demo/settings.py
```diff
@@ -62,7 +62,8 @@
                 'django.template.context_processors.debug',
                 'django.template.context_processors.request',
                 'django.template.context_processors.i18n',
-                'directory_components.context_processors.urls_processor',
+                ('directory_components.context_processors'
+                    '.header_footer_processor'),
```
richtier: do we also need to include the new `urls_processor`?

eadpearce: it's not currently used in the demo. check comments in `context_processors.py` - `header_footer_processor` is just for header/footer templates, `urls_processor` is for when we want to link to services anywhere else

directory_components/context_processors.py
eadpearce: yes good idea 👍

directory_components/context_processors.py
eadpearce: oh! didn't know this was a thing. will do

PR #18, opened 2018-02-26
ED-3543 Fixed incorrect urls

eadpearce:
https://uktrade.atlassian.net/browse/ED-3543


directory_components/version.py
```diff
@@ -1 +1 @@
-__version__ = '1.0.2'
+__version__ = '1.0.3'
```
sdonk: technically this should be version 1.1.0, but don't worry changing it

PR #21, opened 2018-03-01
ED-3615 Added error page title, moved template

eadpearce:
https://uktrade.atlassian.net/browse/ED-3615


demo/settings.py
```diff
@@ -55,7 +55,7 @@
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [],
+        'DIRS': ['directory_components/templates'],
```
richtier: I'm surprised this is needed, because django should look for a folder called `templates` all apps defined by `INSTALLED_APPS`. What happens when you remove this?

eadpearce: aha yes it does actually work without it. i'll remove

PR #24, opened 2018-03-14
ED-3691 widgets demo

eadpearce:
https://uktrade.atlassian.net/browse/ED-3691

screen shot 2018-03-14 at 14 05 38


demo/views.py
```diff
@@ -7,3 +9,8 @@ class HelloWorld(TemplateView):
 
 class NotFound(TemplateView):
     template_name = 'demo/404.html'
+
+
+def demo_form(request):
```
richtier: this can be a `FormView`

```
from django.views.generic.edit import FormView

class NotFound(FormView):
     template_name = 'demo/widgets.html'
     form_class = DemoForm
```

PR #25, opened 2018-03-15
ED-3490 No JS Header

eadpearce:
https://uktrade.atlassian.net/browse/ED-3490

default:
screen shot 2018-03-15 at 13 16 51

with beta banner and international link:
screen shot 2018-03-15 at 13 21 15

mobile: 
localhost_9000_ galaxy s5


demo/requirements.txt
```diff
@@ -1 +1 @@
-export-elements==0.9.1
+export-elements==0.10.0
```
richtier: should this be `0.10.1`?

PR #40, opened 2018-04-19
Fix missing use_required_attribute in CheckboxSelectInlineLabelMultiple

eadpearce:
This fixes the bug in FAB when updating to the latest version of components where all checkboxes in country selector are required


directory_components/widgets.py
```diff
@@ -69,19 +69,17 @@ def get_context(self, *args, **kwargs):
         context['help_text'] = self.help_text
         return context
 
+# change to inherit from ComponentsChoiceWidget instead of django's
+CheckboxSelectMultiple = widgets.CheckboxSelectMultiple
+CheckboxSelectMultiple.__bases__ = (ComponentsChoiceWidget,)
```
konradko: what is this evildoing, why do we need it?

PR #43, opened 2018-05-10
Add components

eadpearce:
screencapture-localhost-9000-components-2018-05-10-14_32_22

related to this ticket https://uktrade.atlassian.net/browse/CMS-39


directory_components/templates/directory_components/components/card.html
```diff
@@ -0,0 +1,13 @@
+<div class="card">
+  <a href="{{ url }}">
+    {% if img %}
+      <img src="{{ img }}" alt="{% if img_alt %}{{ img_alt }}{% endif %}">
```
richtier: `img_src` would be clearer to readers what the variable does.

directory_components/templates/directory_components/components/hero.html
```diff
@@ -0,0 +1,9 @@
+<section class="hero hero-generic"{% if background_image %} style="background-image: url({{ background_image }})"{% endif%}>
```
richtier: similarly, `background_image_url` would be more explanatory

directory_components/templates/directory_components/components/hero.html
```diff
@@ -0,0 +1,9 @@
...
+  <div class="container">
+    <div class="hero-title{% if not large_title %}-compact{% endif %}">
```
richtier: for reference, this could also be `{{ large_title|yesno:'hero-title,hero-title-compact' }}`, but this is fine :)

richtier: actually as you're using the same logic in two places, I think you can do
```
{% with large_title|yesno:'hero-title,hero-title-compact as class_name %}     
```

and then you can use `{{ class_name }}` in two places

directory_components/templates/directory_components/components/social_share_links.html
```diff
@@ -3,25 +3,25 @@
 <ul class="sharing-links" data-component="sharing-links">
     <span class="label">Share</span>
     <li>
-        <a id="share-twitter" rel="noopener noreferrer" target="_blank" href="{{ social_links.twitter|safe }}">
+        <a id="share-twitter" rel="noopener noreferrer" target="_blank" href="{{ twitter|safe }}">
```
richtier: no need for `|safe` here as the url will just be plain text rather than html

directory_components/templates/directory_components/base.html
```diff
@@ -24,8 +24,8 @@
         {% endblock %}
 
         {% block head_css %}
-            <link href="{% static 'export_elements/stylesheets/govuk-elements-styles.min.css' %}" media="all" rel="stylesheet" />
-            <link href="{% static 'export_elements/stylesheets/elements-components.min.css' %}" media="all" rel="stylesheet" />
+            <link href="http://localhost:3000/public/stylesheets/govuk-elements-styles.min.css" media="all" rel="stylesheet" />
+            <link href="http://localhost:3000/public/stylesheets/elements-components.min.css" media="all" rel="stylesheet" />
```
richtier: this will break serving the css on non-local environments such as heroku.

What problem is this solving? or was this not meant to be committed?

directory_components/templates/directory_components/card.html
```diff
@@ -0,0 +1,13 @@
+<div class="card">
+  <a href="{{ url }}">
+    {% if img_src %}
+      <img src="{{ img_src }}" alt="{% if img_alt %}{{ img_alt }}{% endif %}">
```
richtier: the if tags are redundant. it can be simply `{% if img_alt %}{{ img_alt }}{% endif %}`

directory_components/templates/directory_components/base.html
eadpearce: yeah this was not meant to be committed. Was using it for developing locally lol

PR #44, opened 2018-05-17
Expose terms and privacy urls in context

richtier:
[part of this task](https://uktrade.atlassian.net/browse/ED-4239)


directory_components/context_processors.py
```diff
@@ -30,6 +33,23 @@ def analytics(request):
     }
 
 
+@lazy
+def lazy_build_url(url_name, path):
+    return urljoin(get_url(url_name), path)
+
+
+TERMS_AND_CONDITIONS_URL = lazy_build_url(
```
richtier: evaluating lazy to facilitate testing: `get_url` retrieves values from settings.py, which we change in tests - so if we created this variable at module level then the variable would be declared before the setting was changed in the test.

PR #65, opened 2018-07-09
Add option to set html lang attribute in base template

eadpearce:
https://uktrade.atlassian.net/browse/CMS-280


directory_components/templates/directory_components/base.html
```diff
@@ -4,7 +4,7 @@
 
 <!DOCTYPE html>
     <!--[if lt IE 9]><html class="lte-ie8" lang="en"><![endif]-->
-    <!--[if gt IE 8]><!--><html lang="en"><!--<![endif]-->
+    <!--[if gt IE 8]><!--><html lang="{% if html_lang_code %}{{ html_lang_code }}{% else %}en{% endif %}"><!--<![endif]-->
```
richtier: we can use django built in feature for gainz here:

```
{% load get_current_language from i18n %}
{% get_current_language as LANGUAGE_CODE %}
    `lang={{ LANGUAGE_CODE }}`

```
[more info](https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#get-current-language)

For posterity, `if...else` can be shorteneded to: `lang={{ html_lang_code|default:'en' }}`

eadpearce: i thought there was a neater way to do it. thanks

eadpearce: i tried the first example and it works in invest but not in FAS. so depending on the app and how it handles translations we can just set `html_lang_code` in a context processor

PR #79, opened 2018-08-09
Fix JS load order

eadpearce:


eadpearce:
closed as this can be fixed in individual repos that use the package

PR #82, opened 2018-08-28
Add classes for h5 and h6 in template tag

eadpearce:
https://uktrade.atlassian.net/browse/CMS-419

this will fix the bug on the performance dashboard notes page where h5s and h6s aren't styled


demo/templates/demo/template_tags.html
```diff
@@ -0,0 +1,49 @@
...
+
+{% block head_title %}directory-components{% endblock %}
+
+{% block content %}
+<section>
+  <div class="container">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+
+        {% filter add_export_elements_classes %}
```
richtier: I did not know about `{% filter ... ` :+1:

PR #100, opened 2018-11-23
Add django admin IP restrictor middleware helper

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-768)

this wrapper around the django admin ip restrictor middleware handles the fact PaaS IP address is retrieved differently than on heroku. Stolen from directory-api

### DIFF
--- a/directory_components/helpers.py
+++ b/directory_components/helpers.py
@@ -1,6 +1,11 @@
 import urllib.parse
 
+from django.conf import settings
 from django.utils.encoding import iri_to_uri
+
+from ipware.ip2 import get_client_ip
+
+from directory_components import constants
 
 
 def add_next(destination_url, current_url):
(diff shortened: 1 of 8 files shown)


tests/urls.py
```diff
@@ -1,10 +1,23 @@
...
+from django.conf.urls import include, url
 from django.views import View
+from django.views.generic.base import RedirectView
 
 import directory_components.views
 
 
+admin_urls = [
+    url(
+        r"^thing/$",
```
kowalcj0: Was ist das? :smile:

kowalcj0: is it a dummy `404` page when requests comes from restricted IP address?

directory_components/helpers.py
```diff
@@ -73,3 +78,71 @@ def full_path(self):
...
+        """
+        if 'HTTP_X_FORWARDED_FOR' not in request.META:
+            raise LookupError(cls.MESSAGE_MISSING_HEADER)
+
+        x_forwarded_for = request.META['HTTP_X_FORWARDED_FOR']
+        ip_addesses = x_forwarded_for.split(',')
+        if len(ip_addesses) < 2:
+            raise LookupError(cls.MESSAGE_INVALID_IP_COUNT)
+
+        return ip_addesses[-2].strip()
```
kowalcj0: arghhh magic number here and in previous `if`. 
Would prefer to have a descriptive variable instead

tests/urls.py
richtier: it's for use in tests :)

it's used in a few tests:

https://github.com/uktrade/directory-components/pull/100/files#diff-da5a345e8216a8e47d28218a4f5b9b0fR240

https://github.com/uktrade/directory-components/pull/100/files#diff-da5a345e8216a8e47d28218a4f5b9b0fR284

directory_components/helpers.py
richtier: :+1:

richtier: oh actually it's explained in the docstring :)
one of the few times I've ever writtena  docstring :laughing:

```py
        """
        Returns the IP of the client making a HTTP request, using the
        second-to-last IP address in the X-Forwarded-For header. This
        should not be able to be spoofed in GovukPaaS, but it is not
        safe to use in other environments.

        Args:
            request (HttpRequest): the incoming Django request object

        Returns:
            str: The IP address of the incoming request

        Raises:
            LookupError: The X-Forwarded-For header is not present, or
            does not contain enough IPs
        """
```

PR #101, opened 2018-11-27
Skip IP check if signature in cookie is valid to facilitate tests

richtier:


directory_components/helpers.py
```diff
@@ -146,3 +148,23 @@ def __new__(cls, *args, **kwargs):
...
+        'id': settings.IP_RESTRICTOR_SKIP_CHECK_SENDER_ID,
+        'key': settings.IP_RESTRICTOR_SKIP_CHECK_SECRET,
+        'algorithm': 'sha256'
+    }
+    try:
+        Receiver(
+            lambda x: credentials,
+            signature,
+            '/',
+            'GET',
```
kowalcj0: cookie checker has to support also: `POST`, `DELETE`, `PUT`, `PATCH` methods

PR #104, opened 2018-11-30
Remove export journey links from header/footer + feature flag

eadpearce:


setup.py
```diff
@@ -34,6 +34,13 @@
             "twine>=1.11.0,<2.0.0",
             "wheel>=0.31.0,<1.0.0",
             "setuptools>=38.6.0,<39.0.0"
+        ],
+        "demo": [
+            "django-environ==0.4.5",
+            "gunicorn==19.5.0",
+            "whitenoise==3.1",
+            "pip==9.0.1",
+            "beautifulsoup4>=4.6.0<5.0.0",
```
richtier: I think this needs to be `>=4.6.0,<5.0.0`

the current way works, but has odd outcomes sometimes

PR #107, opened 2018-12-05
Allow paas ip retriever to return 2 ips

sdonk:


directory_components/helpers.py
```diff
@@ -118,6 +120,8 @@ def get_ip_address(cls, request):
             LookupError: The X-Forwarded-For header is not present, or
             does not contain enough IPs
         """
+        IPs = namedtuple('IPs', ['second', 'third'])
```
richtier: could this be moved to module level?

directory_components/helpers.py
```diff
@@ -126,10 +130,14 @@ def get_ip_address(cls, request):
         if len(ip_addesses) < 2:
             raise LookupError(cls.MESSAGE_INVALID_IP_COUNT)
 
-        return ip_addesses[-2].strip()
+        if len(ip_addesses) == 3:
```
richtier: if cloudfront config changes the number of IPs may increase.

I think it's safer to check if there is at least three.

PR #126, opened 2019-03-20
Feature flag new header

eadpearce:


directory_components/templates/directory_components/header_footer/domestic_header.html
```diff
@@ -71,8 +67,23 @@
...
 </li>
+{% endif %}
+
+{% endblock %}
 
+{% block header_search %}
+<div class="great-header-search-wrapper">
+  <div class="great-header-search-container">
+    {% block header_search_form %}
+      <form action="{{ services_urls.exred }}/search" method="GET">
```
richtier: `/search/` trailing slash pls

tests/test_header_footer.py
```diff
@@ -642,19 +667,50 @@ def test_service_header_adjusted_width_according_to_market_access_feature(
...
+))
+def test_new_header_footer_feature_flag_off(template_name, type, settings):
+    context = {
+        'features': {'NEW_HEADER_FOOTER_ON': False}
+    }
+    html = render_to_string(template_name, context)
+
+    soup = BeautifulSoup(html, 'html.parser')
+
+    assert soup.find(id='{}-dit-logo'.format(type))
```
richtier: `type` is a built-in python object. please rename to e.g., `element_type`

setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='directory_components',
-    version='6.1.0',
+    version='6.2.0',
```
richtier: is this not a breaking change?

eadpearce: possibly? not sure. but it makes sense for this to be major version anyway

PR #128, opened 2019-03-20
Don't display caption when an image-with-caption doesn't have one

Kenneth-Siu:
Trying to avoid this:

image


directory_components/templates/directory_components/image_with_caption.html
```diff
@@ -1,7 +1,9 @@
 <figure class="image-with-caption">
   <img src="{% firstof img_url image_url image_src img_src image %}" alt="{{ alt_text }}">
+  {% if main_caption or sub_caption %}
```
eadpearce: this can just check for `main_caption` since `sub_caption` is under an if block already but main isn't

Kenneth-Siu: If someone adds a sub_caption but not a main_caption, then nothing will turn up which is weird.

PR #130, opened 2019-03-21
Fix css path in base template

eadpearce:


directory_components/templates/directory_components/base.html
```diff
@@ -39,7 +39,7 @@
             .js-disabled-only { display: none; }
         </style>
         <noscript>
-          <link href="{% static 'export_elements/stylesheets/elements-components-no-js.min.css' %}" media="all" rel="stylesheet" />
+          <link href="{% static 'directory_components/export_elements/stylesheets/elements-components-no-js.min.css' %}" media="all" rel="stylesheet" />
```
Kenneth-Siu: Does this file get checked in to the repository? I can't see it here... https://github.com/uktrade/directory-components/tree/master/directory_components/static/directory_components/export_elements/stylesheets

eadpearce: AH. good spot. we're missing a gulp task. i can fix that

PR #131, opened 2019-03-22
Fix cookie name default + bug fixes

eadpearce:


MichaelWalker:
Happy with the COOKIE stuff too! thanks Edie!

PR #135, opened 2019-03-25
Statistics component

Kenneth-Siu:
The section in the demo looks a bit like this but with better spacing and nicer font:

image


demo/templates/demo/cards.html
```diff
@@ -42,6 +42,13 @@ <h2 class="heading-medium">CTA card</h2>
         <pre><code>{% verbatim %}{% cta_card with_arrow=True image_url=card_image02 text="Tell us what help you need" %}{% endverbatim %}</code></pre>
       </div>
     </div>
+    <div>
+      <h2 class="heading-medium">Statistics</h2>
+      <p>A grid of statistics</p>
+      {% statistics_card_grid statistics=view.statistics num_of_statistics=view.num_of_statistics %}
```
richtier: grid implies both columns and rows. this component looks more like a row

(I'm being influenced by the new "grid" CSS layout)

`statistics_card_row` seems more appropriate

richtier:
how does it look on smaller screens?

demo/templates/demo/cards.html
Kenneth-Siu: The main thing with the naming was that it uses a card-grid for its formatting and I wanted the name to make it explicit that it's a card grid and not something you're meant to put inside the card-grid.

Kenneth-Siu:
> how does it look on smaller screens?

image
image
image

(With better spacing. npm issues, don't go into it. We know it works because it's porting existing code from other projects)

PR #139, opened 2019-03-27
CMS-1038 — Design review

Kenneth-Siu:


directory_components/templates/directory_components/cta_card.html
```diff
@@ -3,6 +3,6 @@
 <div class="cta-card">
   <img src="{% firstof img_url image_url image_src img_src image %}" alt="{{ alt_text }}">
   <div class="link-container">
-    <a class="cta-link {% if with_arrow %}with-arrow{% endif %}" href="{{ url }}">{{ text|split_lines:25 }}</a>
+    <a class="cta-link {% if with_arrow %}with-arrow{% endif %}" href="{{ url }}"><span>{{ text }}</span><br/></a>
```
richtier: are the spans adding value here?

directory_components/templatetags/directory_components_tags.py
```diff
@@ -19,13 +19,6 @@ def static_absolute(parser, token):
     return FullStaticNode.handle_token(parser, token)
 
 
-@register.filter
-def split_lines(text, length):
```
richtier: removing a feature makes this a breaking change, so please change version by a major version to communicate that (so version 8.0.0)

directory_components/templates/directory_components/cta_card.html
Kenneth-Siu: Yep! The arrow (and :after in the scss) shouldn't have a blue background, but the text does, so the span is there to give only the text a blue background.

directory_components/templatetags/directory_components_tags.py
Kenneth-Siu: Actually, speaking of, @richtier, any opinions on removing something like this? I know it's not used anywhere within code in uktrade (which is why I removed it), but I'm happy to keep it if that's the preference.

richtier: if it's not used then yes by all means please remove :+1:

PR #152, opened 2019-04-04
Hero banner added

jessgilbert:
Hero banner added that's used for invest home page, international home page and the new international capital invest home page.
image
image
image


directory_components/export_elements/sass/components/hero/_base.scss
```diff
@@ -2,6 +2,16 @@
 @import '../partials/spacings';
 
 // Hero banner
+$spacing-1: 60px;
+$spacing-2: 45px;
+$spacing-3: 30px;
+$spacing-4: 15px;
+$white: #ffffff;
```
eadpearce: these vars should be in `../partials/colours` you can just import them

directory_components/export_elements/sass/components/hero/_base.scss
```diff
@@ -89,3 +99,69 @@
...
+
+.hero-with-cta {
+    padding: $spacing-1 + $spacing-3 0 $spacing-1;
+    background-color: $black;
+
+    h1 {
+      margin-top: 0;
+      margin-bottom: $spacing-4;
+      font-family: $hero-brand-font;
+      text-transform: uppercase;
```
eadpearce: check with michal but we should be using non-uppercase for all heroes except for the one on domestic that has the red highlight (brand thing)

directory_components/export_elements/sass/components/hero/_base.scss
```diff
@@ -2,6 +2,16 @@
 @import '../partials/spacings';
 
 // Hero banner
+$spacing-1: 60px;
```
richtier: is it worth reusing these? https://github.com/uktrade/directory-components/blob/master/directory_components/export_elements/sass/partials/_spacings.scss

directory_components/templates/directory_components/hero_with_cta.html
```diff
@@ -0,0 +1,17 @@
+{% load static %}
+<section class="hero hero-with-cta" style="background-image: url('{{ img_url }}')">
+	<div class="container">
+		<div class="hero-text">
+			<h1 class="heading-xlarge margin-bottom-15">{{ heading }}</h1>
```
eadpearce: we can't have 2 `h1`s on a page. pls combine these and use a `div` or `span` to separate the heading and subheadings. or use an `h1` and an `h2`

directory_components/templates/directory_components/hero_with_cta.html
```diff
@@ -0,0 +1,17 @@
+{% load static %}
```
richtier: as far as I can tell this is not used

for future when static is used can it please be {% load x from static %} as without it, it's analogous to `from foo import *`, which is hard for a reader to understand what's being used and where it's coming from :)

PR #154, opened 2019-04-08
Removed linear gradient

jessgilbert:
image
I added this as I needed a gradient for the specific case I was using it, however having it on all is not appropriate I don't think and I have added the gradient to the actual image as it is required every time the image is used


jessgilbert:
> please update setup.py version :)

jessgilbert:
Added padding top and bottom rather than min-height and image is picked to fit how it's wanted on page and so min-height is coming out too large in cases

PR #155, opened 2019-04-08
CMS-1165 — Header language select

Kenneth-Siu:
image


directory_components/helpers.py
```diff
@@ -195,3 +196,19 @@ def get_country_from_querystring(request):
 def get_user_country(request):
     return get_country_from_querystring(request) or \
         request.COOKIES.get(constants.COUNTRY_COOKIE_NAME, '')
+
+
+def get_language_from_prefix(path):
```
eadpearce: i think you can remove these two functions - i copied them in from invest but they're not actually used anywhere

PR #163, opened 2019-04-10
Language selector can turn off

Kenneth-Siu:


directory_components/templates/directory_components/header_footer/international_header.html
```diff
@@ -38,6 +38,7 @@
 </div>
 {% endif %}
 
+{% if hide_language_switcher %}
```
richtier: should this be `{% it not hide_language_switcher %}` ?

Kenneth-Siu: lol yes

PR #164, opened 2019-04-11
Button only shows when a link and text are provided

jessgilbert:


directory_components/templates/directory_components/hero_with_cta.html
```diff
@@ -8,7 +8,7 @@ <h2 class="heading-xlarge margin-top-0 margin-bottom-30">{{ subheading }}</h2>
 			{% if subtitle %}
 			<p class="hero-subtitle margin-bottom-45">{{ subtitle }}</p>
 			{% endif %}
-			{% if cta_text %}
+			{% if cta_text and cta_link %}
 				<a href={{cta_link}} class="button button-arrow-small button-blue margin-bottom-30">{{ cta_text }}</a>
```
richtier: `href="{{cta_link}}"` :)

richtier: perhaps pedantic, but as you're going there anyway, could you please also do `{{ cta_link }}` (with spaces at start and end) to be consistent with prevailing template variable style :)

PR #167, opened 2019-04-11
Allow use of lang or language for form field name

eadpearce:


directory_components/templates/directory_components/header_footer/header_base.html
```diff
@@ -12,7 +12,7 @@
       {% block global_header_logo %}
       {% endblock %}
       <div class="great-domestic-international-links">
-        <a href="{{ services_urls.great_domestic }}" id="great-global-header-domestic-link" class="domestic">{% trans 'For UK businesses' %}</a>
+        <a href="{{ services_urls.great_domestic }}?lang=en-gb" id="great-global-header-domestic-link" class="domestic">{% trans 'For UK businesses' %}</a>
```
richtier: this would change the user back to english. is this desirable?

eadpearce: yeah if we don't explicitly set the language when they go to domestic homepage the geoip kicks in and boots them back to /international (or at least, this has happened to me before)

tests/test_forms.py
```diff
@@ -1,10 +1,17 @@
...
 def test_get_language_form_initial_data():
     translation.activate('fr')
     request_stub = {}
-    data = get_language_form_initial_data(request_stub)
+    data = forms.get_language_form_initial_data(request_stub)
     assert data['language'] == 'fr'
+
+
+def test_get_lang_form_initial_data():
+    translation.activate('fr')
```
richtier: this change affect other tests/code that use `translation.get_language()` that run after this test
we can avoid that by doing 

```
with translation.override('fr'):
    request_stub = {}
    data = forms.get_lang_form_initial_data(request_stub)
    assert data['lang'] == 'fr'
    
```

eadpearce: ooh neat. didn't realise you could do that

PR #172, opened 2019-04-16
Include compiled translations

richtier:


demo/config/settings.py
```diff
@@ -111,7 +111,7 @@
 ]
 
 LOCALE_PATHS = (
-    os.path.join(BASE_DIR, 'locale'),
+    os.path.abspath(os.path.join(BASE_DIR, '../directory_components/locale')),
```
eadpearce: couldn't you just do `PROJECT_ROOT, 'directory_components/locale'`

richtier: oddly, `PROJECT_DIR` is `directory-components/demo/config/`

PR #178, opened 2019-04-25
Add breadcrumbs

richtier:
image


demo/templates/demo/breadcrumbs.html
```diff
@@ -0,0 +1,44 @@
...
+        <h2 class="heading-medium">Three levels</h2>
+        <pre><code>{% verbatim %}{% breadcrumbs left_url='http://www.example.com' left_label='home' middle_url='http://www.example.com/middle/' middle_label='middle label' right_label='right label' %}{% endverbatim %}</code></pre>
+        {% breadcrumbs left_url='http://www.example.com' left_label='home' middle_url='http://www.example.com/middle/' middle_label='middle label' right_label='right label' %}
+      </div>
+    </div>
+    <div class="grid-row"> 
+      <div class="column-full padding-bottom-15">
+        <h2 class="heading-medium">Two levels</h2>
+        <p>Similar to three levels, but exclude middle_url and middle_label.</p>
+        <pre><code>{% verbatim %}{% breadcrumbs left_url='http://www.example.com' left_label='home' middle_url='http://www.example.com/middle/' middle_label='middle label' right_label='right label' %}{% endverbatim %}</code></pre>
```
MichaelWalker: This still contains the 'middle_label' and 'middle_url' attributes.
Should they be excluded from this fragment?

directory_components/templates/directory_components/breadcrumbs_flat.html
```diff
@@ -0,0 +1,13 @@
+<div class="breadcrumbs">
+  <ol>
+    <li>
+      <a href="{{ left_url }}">{{ left_label }}</a>
+    </li>
+    {% if middle_label %}
+	    <li>
+	      <a href="{{ middle_url }}">{{ middle_label }}</a>
+	    </li>
+	{% endif %}
```
MichaelWalker: slightly odd indentation here if we are being super picky! 😊

directory_components/templates/directory_components/breadcrumbs_flat.html
```diff
@@ -0,0 +1,13 @@
+<div class="breadcrumbs">
```
MichaelWalker: Do we ever have more that 3 levels to handle? 
I can't think of an example of where we do, but don't know the site well enough to be sure yet!

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+
+## 10.6.0
```
MichaelWalker: Possibly missing something obvious here, but I don't see where the version has been updated.

MichaelWalker: Also this is now slightly out of date - current version of directory components is now 10.7.0

demo/templates/demo/breadcrumbs.html
richtier: thanks. yep :)

directory_components/templates/directory_components/breadcrumbs_flat.html
richtier: tl;dr you're correct. I'll change stuff.

we used to use `breadcrumbs.html`, which supports multi levels. I think it was used by CMS pages. Looks like it's not used anwyere now https://github.com/search?l=HTML&q=org%3Auktrade+directory_components%2Fbreadcrumbs.html&type=Code

will simplify this stuff now

CHANGELOG.md
richtier: I think I made this PR before another PR was made and merged which also used 10.6.0...maybe

will use 10.8.0 :)

PR #182, opened 2019-04-29
Add GA tagging file for international header.

MichaelWalker:
Contains the tags that can be used to track behaviour of the international header.

We are going to want to add tags to the header in lots of different repos, so adding it here as a shared place.

@eadpearce @agiamas @stevenburnell-dit - I couldn't find any up to date information on the spreadsheet that related to the new international header, so this is a best guess at what I think we should be tracking. At the moment, I'm tracking the country changer, the language changer, and the navigation links from the header. Let me know if I've missed anything or if anything needs changing!

I'll update the spreadsheet with what I've done once we are happy with this review.

I'm not sure how to be confident that this will all work as I'd like it to, (short of merging it in, using it in a project and getting it on invest). If anyone has a better plan (ideally a nice way of testing it works locally) then I'd love any suggestions!


directory_components/static/directory_components/js/dit.tagging.internatinalHeader.js
```diff
@@ -0,0 +1,64 @@
+dit.tagging.internationalHeader = function() {
+
+    this.init = function() {
+        $("[data-ga-id='choose-country-form']").on("submit", function() {
```
MichaelWalker: I've opted to use data attributes as the selector as it makes everything nice and explicit, and it reduces the chances of accidentally messing up the logging when moving elements around / re-styling things.

Open to other suggestions though! :)

directory_components/static/directory_components/js/dit.tagging.internatinalHeader.js
```diff
@@ -0,0 +1,64 @@
+dit.tagging.internationalHeader = function() {
+
+    this.init = function() {
+        $("[data-ga-id='choose-country-form']").on("submit", function() {
+            window.dataLayer.push({
+                'eventAction': 'Submit',
+                'eventCategory': 'Select Country',
+                'eventLabel': 'International Header',
+                'eventValue': $("#great-header-country-select").val()
```
MichaelWalker: @stevenburnell-dit - hopefully I've got roughly the right kind of convensions here for the event data.

Let me know if anything can be changed to bring it more in line with what you've already done!

stevenburnell-dit: While there isn't a definitive list to follow there is no right or wrong on this, right now. The code does the right thing in attempting to resemble tagging values we've used elsewhere, so I'm inclined to say it's fine.

If I was to be picky, I might consider a small change to this (yes, eventCategory omitted):
window.dataLayer.push({
                'eventAction': 'Select Country',
                'eventLabel': 'International Header',
                'eventValue': $("#great-header-country-select").val()

That's only because I've tried to use 'eventAction' elsewhere as an identifier for a specific user action/something the user does. My (unexplained) thoughts imagining we might want to collectively record such actions wherever they occur and use the data to find out the most popular method (e.g. user changed the country by International header rather than using blah link somewhere else).

directory_components/static/directory_components/js/dit.tagging.internatinalHeader.js
```diff
@@ -0,0 +1,64 @@
...
+            window.dataLayer.push(navLinkClickedEvent("How to do business with the UK"));
+        });
+
+        $("[data-ga-id='news-header-navigation']").on("click", function() {
+            window.dataLayer.push(navLinkClickedEvent("News"));
+        });
+
+        $("[data-ga-id='domestic-header-navigation']").on("click", function() {
+            window.dataLayer.push(navLinkClickedEvent("For UK businesses"));
+        });
```
stevenburnell-dit: Looks like an opportunity to reduce code here.

Is there an advantage/reason for multiple (single item only??) navigation selectors or, could the code be reduced to something like:
$("[data-ga-class='international-header-navigation']").on("click", function() {
            window.dataLayer.push({
                'eventAction': 'Navigation',
                'eventCategory': 'Header Link',
                'eventLabel': 'International Header',
                'eventValue': $(this).text()
            });
        });

Note: 
data-ga-class only suggested to indicate similar 'multiple' idea seen in CSS
I'd have no problem with keeping data-ga-id

MichaelWalker: Good idea, thanks. I'll make that change.

directory_components/templates/directory_components/header_footer/international_header.html
```diff
@@ -59,26 +59,26 @@
...
 <li class="column-quarter js-menuitem">
-  <a href="{{ header_footer_urls.how_to_do_business_with_the_uk }}" class="link-heading" id="header-how-to-do-business-with-the-uk">{% trans 'How to do business with the UK' %}</a>
+  <a href="{{ header_footer_urls.how_to_do_business_with_the_uk }}" class="link-heading" id="header-how-to-do-business-with-the-uk" data-ga-id="how-to-do-business-header-navigation">{% trans 'How to do business with the UK' %}</a>
 </li>
 {% endif %}
 
 {% if features.NEWS_SECTION_ON %}
 <li class="column-quarter js-menuitem">
-  <a href="{{ header_footer_urls.international_news }}" class="link-heading" id="header-news">{% trans 'News' %}</a>
+  <a href="{{ header_footer_urls.international_news }}" class="link-heading" id="header-news" data-ga-id="domestic-header-navigation">{% trans 'News' %}</a>
```
eadpearce: should this be `international-news-header-navigation` instead?

directory_components/templates/directory_components/header_footer/header_country_select.html
```diff
@@ -6,7 +6,7 @@
     <div class="countries">
       <h2>Choose your country</h2>
       <p>Tell us where your business is based to see information, guidance and opportunities specific to your location.</p>
-      <form class="form-group" action="" method="GET">
+      <form class="form-group" action="" method="GET" data-ga-id="choose-country-form">
```
eadpearce: maybe `banner-change-country-form` would make this easier to differentiate from the `change-country-form` in the header

directory_components/templates/directory_components/header_footer/international_header.html
MichaelWalker: good spot! Yes, (but changed to data-ga-class='header-navigation' anyway now as per Steven's good suggestion.

directory_components/templates/directory_components/header_footer/header_country_select.html
MichaelWalker: nice! Thanks

PR #184, opened 2019-05-02
Cms 1395 transition from old cookie name

Kenneth-Siu:


directory_components/middleware.py
```diff
@@ -115,6 +115,10 @@ def process_request(self, request):
 
 class PersistLocaleMiddleware:
     def process_response(self, request, response):
+        if hasattr(settings, 'LANGUAGE_COOKIE_DEPRECATED_NAME'):
+            response.delete_cookie(
```
richtier: why delete the old cookie?

if we change the name then the old one will be benign

so it seems we would be adding more logic and maintenance overhead with no gain

eadpearce: because django says so :D 
> Be cautious when updating this setting on a production site. If you update this setting to use a deeper path than it previously used, existing user cookies that have the old path will not be updated. This will result in site users being unable to switch the language as long as these cookies persist. The only safe and reliable option to perform the switch is to change the language cookie name permanently (via the LANGUAGE_COOKIE_NAME setting), and to add a middleware that copies the value from the old cookie to a new one and then deletes the one.

richtier: if the docs say jump then I say "how high" :laughing:

PR #194, opened 2019-05-08
Stop language/country cut-off in header in IE, Edge, and FF

Kenneth-Siu:


Kenneth-Siu:
> Please update the changelog

Derp

PR #196, opened 2019-05-14
Explicitly set text direction in header country select banner

eadpearce:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 12.0.1
```
sdonk: This is now 12.2.0

setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='directory_components',
-    version='12.0.0',
+    version='12.0.1',
```
sdonk: ditto

PR #197, opened 2019-05-15
CI-108: Update and enforce page level GA tagging to match new spec.

MichaelWalker:


directory_components/mixins.py
```diff
@@ -91,10 +91,36 @@ def get_context_data(self, *args, **kwargs):
...
+        self.business_unit = business_unit
+        self.site_section = site_section
+        self.site_subsection = site_subsection
+        # site language, user_id and login_status are all set automatically
+        # within the GA360Mixin
+
+
 class GA360Mixin:
-    ga360_payload = None
+    ga360_payload = None  # This field should be set in the views.
```
MichaelWalker: Am I missing a more pythonic way of doing this. (using the class seems a little clunky)

We could just use a dictionary, but I want to make it apparent what fields we should be expecting (hence the class)

MichaelWalker: Actually a `set_ga_payload(page_id, business_unit, site_section, site_subsection=None)` method within the mixin would be a little nicer.

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
...
+
+        # Don't check views which should be skipped (see @skip_ga360 decorator)
+        if getattr(response, 'skip_ga360', False):
+            return response
+
+        if not hasattr(response, 'context_data'):
+            raise GADataMissingException('No context data found')
+        context_data = response.context_data
+
+        if 'ga360' not in context_data:
```
eadpearce: this throws a 500 on any view if the GA data is missing and we haven't explicitly skipped it on the view. is this desirable?

MichaelWalker: Yep! @richtier  and I had a fairly long conversation about exactly this! 😊

The argument is that Analytics is actually business critical, and we wish for changes that don't think about it to break in big and obvious ways before they get anywhere near production.

It obviously makes the initial release of any app updating to use this version a little scary, but we've got some good tools for checking all pages and making ourselves confident that nothing is broken, so hopefully this should also be OK.

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
     def process_exception(self, request, exception):
         if hasattr(request, 'LANGUAGE_CODE') and request.LANGUAGE_CODE:
             translation.activate(request.LANGUAGE_CODE)
+
+
+class GADataMissingException(Exception):
+    def __init__(self, message):
+        super(GADataMissingException, self).__init__(message)
```
richtier: would shortcut of `super().__init__(message) suffice?

in fact, is this needed :)

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
...
+
+
+Field = namedtuple('Field', ['name', 'allowed_types'])
```
richtier: new python feature that substitutes `namedtuple` for this uscase (just FYI in case you missed it):

https://docs.python.org/3/library/dataclasses.html 
```
@dataclass
class Field:
    name: str
    allowed_types: list
```

up to you if you want to use it, of course :)

richtier: new python feature that substitutes `namedtuple` for this uscase (just FYI in case you missed it):

https://docs.python.org/3/library/dataclasses.html 
```
@dataclass
class Field:
    name: str
    allowed_types: list
```

up to you if you want to use it, of course :)

directory_components/middleware.py
MichaelWalker: Good point - I'll update

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
...
+required_fields = [
+    Field(name='business_unit', allowed_types=str),
+    Field(name='site_section', allowed_types=str),
+    Field(name='user_id', allowed_types=(str, type(None))),
+    Field(name='login_status', allowed_types=bool),
+    Field(name='site_language', allowed_types=str)
```
richtier: could this be a consistent list, please, so `allowed_types=[str]` - less things to think about if an attribute is consistently the same type

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
...
+class GADataMissingException(Exception):
+    def __init__(self, message):
+        super(GADataMissingException, self).__init__(message)
+
+
+Field = namedtuple('Field', ['name', 'allowed_types'])
+required_fields = [
+    Field(name='business_unit', allowed_types=str),
+    Field(name='site_section', allowed_types=str),
+    Field(name='user_id', allowed_types=(str, type(None))),
```
richtier: I think `type(None) == None`, so this could be `allowed_types=[str, None])`?

richtier: TIL we can't do `isistance(foo, None)` :)

directory_components/middleware.py
MichaelWalker: I had missed it! Nice

This would force a dependency on python 3.7 or greater though right, which might break things.
I think I'll leave it for now, to make us more forgiving on python version.

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
...
+    Field(name='login_status', allowed_types=bool),
+    Field(name='site_language', allowed_types=str)
+]
+
+
+class CheckGATags:
+    def process_response(self, _, response):
+
+        # Only check 2xx responses for google analytics.
+        if response.status_code < 200 or response.status_code > 299:
```
richtier: sugar: `200 <= response.status_code < 300:`

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
...
+            return response
+
+        # Don't check views which should be skipped (see @skip_ga360 decorator)
+        if getattr(response, 'skip_ga360', False):
+            return response
+
+        if not hasattr(response, 'context_data'):
```
richtier: how could `response.context_data` not have context_data? redirects would not have context data but that would be handled by the `if 200` above

directory_components/middleware.py
richtier: indeed, I'm under the impression GA data is a first-class citizen. If we detected in prod it was missing then we would need to fix it. raising an error in the view highlights the error early (so it can be caught in dev or stage).

@agiamas are to comment?

agiamas: That sounds about right, we should have GA data for the pages that we can and know that we dont want to for the ones that we cant.

directory_components/middleware.py
```diff
@@ -148,3 +149,55 @@ def process_response(self, request, response):
...
+            raise GADataMissingException('No context data found')
+        context_data = response.context_data
+
+        if 'ga360' not in context_data:
+            raise GADataMissingException(
+                "No Google Analytics data found on the response.")
+
+        ga_data = context_data['ga360']
+        for field in required_fields:
```
richtier: [JSON schema validation](https://richardtier.com/2014/03/24/json-schema-validation-with-django-rest-framework/) might simplify this task.

the blog post is in context of validating an endpoint, but there is transferable code :)

directory_components/middleware.py
MichaelWalker: can't be a list (which is why we ended up like this) as `instance` won't accept that as an argument, but I've just found the syntax for single element tuples, which is slightly forced but less bad than the different types! I'll update.

directory_components/mixins.py
```diff
@@ -92,9 +92,26 @@ def get_context_data(self, *args, **kwargs):
 
 
 class GA360Mixin:
-    ga360_payload = None
+    ga360_payload = {}
+
+    def set_ga360_payload(self, page_id, business_unit, site_section,
```
richtier: I love syntactic sugar, and could not help notice this could be:

```
    @property
    def ga360_payload(self):
         ...

    @ga360_payload.setter
    def ga360_payload(self, value):
        .... # the logic currently in `set_ga360_payload`

...

self.ga360_payload = {
    'page_id': 'TestPageId',
    'business_unit': 'Test App',
    'site_section': 'Test Section',
    'site_subsection': 'Test Page'
)
```

your fall. just :heart: sugar.

directory_components/middleware.py
MichaelWalker: I wasn't confident no 200 response ever returns without context data.
What about things like `sitemap.xml` etc?

If its true that this can't ever be null at this point then let me know and I'll happily change it.

directory_components/middleware.py
richtier: > New in version 3.7.

oh damn you're right :) one day we can use. and it will be amazing.

directory_components/middleware.py
richtier: coolio :)

```
Field(name='site_language', allowed_types=(str,))
```

directory_components/middleware.py
richtier: `sitemap.xml` is just a TemplateView really, so that has `context_data` :)

https://github.com/django/django/blob/master/django/contrib/sitemaps/views.py#L44

directory_components/mixins.py
MichaelWalker: nice! I played around with this a bit, but ended up picking the method as it makes the field names that we are expecting part of the public API for the class.

I agree that the setter makes code that is nicer to read/write though, so very open to being pursuaded! :)

richtier: lets see if this can push you over the edge:

- passing in a dict means less unpacking
- if you using schema validation that expects a dict anyway

it's your call tho.

MichaelWalker: (as discussed offline)

The validation isn't done here (as then we'd miss if a view never called the setter) so the setter wouldn't need to exist.

I'm going to leave it as it is, as otherwise it would be unclear to another dev which properties should be passed into this method.

PR #199, opened 2019-05-23
Remove unused old headers/footers and feature flags to do with them

Kenneth-Siu:


CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
 # Changelog
 
+## 14.0.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/199/files)
+
+### Breaking Changes
+
+- Feature flag no longer used:
+    - `NEW_HEADER_FOOTER_ON`
+- Six templates have been removed:
```
eadpearce: 🔥

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+    - `header_footer_old/header.html`
+    - `header_footer_old/header_international_link_js_disabled.html`
+    - `header_footer_old/header_static.html`
+- Images removed:
+    - `IIG_Invest_in_GREAT_White_WITH_Flag.png`
+    - `UK-Gov_STACK_WHITE_AW.png`
+- JavaScript files removed:
+    - `invest-header.js`
+- Templates renamed/moved:
+    - `header_footer/header_static.html` changed to `header_footer/domestic_header_static.html`
```
eadpearce: if these templates are kept with the same name as before it will make for fewer changes when we upgrade the library in services. BUT this does make naming more explicit... @richtier how do you feel about this

Kenneth-Siu: I checked and the number of times these renamed/moved files are used can be counted with one hand.

(Now that I think of it, the only one I remember finding when doing a search was in SOO)

PR #204, opened 2019-05-29
CI-108: Fix GA360 mixin failing when request.sso_user is unset.

MichaelWalker:


directory_components/mixins.py
```diff
@@ -105,7 +105,8 @@ def set_ga360_payload(self, page_id, business_unit, site_section,
     def get_context_data(self, *args, **kwargs):
         self.ga360_payload['site_language'] = translation.get_language()
 
-        if self.request.sso_user and self.request.sso_user.id:
+        if (hasattr(self.request, 'sso_user')
+                and self.request.sso_user and self.request.sso_user.id):
```
richtier: would this be cromulent? 

```
try:
    self.ga360_payload['user_id'] = str(self.request.sso_user.id)
    self.ga360_payload['login_status'] = True
except AttreibuteError:
    self.ga360_payload['user_id'] = None
    self.ga360_payload['login_status'] = False
```

MichaelWalker: yes - that's nicer!

PR #205, opened 2019-05-29
CI-145 Added invest_header using invest logo specifically for invest pages

Kenneth-Siu:


demo/templates/demo/invest-header.html
```diff
@@ -0,0 +1,42 @@
+{% extends 'demo/base.html' %}
+{% load static %}
+{% load breadcrumbs from directory_components_tags %}
```
jessgilbert: Is this now just meant to be `directory_components` ?

PR #215, opened 2019-06-11
[CMS 1507] Add Django 2 -> 2.2 compatibility

ababic:
These changes should maintain compatibility with Django 1.11, whilst also supporting apps using Django versions up to 2.2.

There was a bit of a dependency-chain to follow (all demo and test requirements), and there were a couple of minor changes needed to keep the demo and test apps working.

The important changes here, in relation to compatibility with other apps, are:
- Middleware is now compatible with all supported Django versions, and can be used in the newer `MIDDLEWARE` setting, or the older `MIDDLEWARE_CLASSES` one.
- The included template tags should also be usable in all supported Django versions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 16.3.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/215/files)
+
+### Implemented enhancements
+- Added support for Django 2 -> 2.2
+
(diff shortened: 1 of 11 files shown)


ababic:
I will merge this soon, AFTER i've updated the CHANGELOG :)

MichaelWalker:
Actually - missed during my review, but you will also need to update the version number in setup.py

PR #216, opened 2019-06-12
Add message box with icon component + documentation

eadpearce:
screencapture-components-trade-great-9013-message-boxes-2019-06-12-13_52_57
Screen Shot 2019-06-12 at 15 17 11

Also added 500 page template because we still didn't have one! Based on the one we use in FAS, tweaked a bit to be more in line with the pattern on [gov.uk design system](https://design-system.service.gov.uk/patterns/problem-with-the-service-pages/)

Screen Shot 2019-06-12 at 15 18 01


directory_components/export_elements/overrides/govuk_elements/elements/_layout.scss
```diff
@@ -59,7 +59,7 @@ section {
 // when only one column is needed
 
 .width-sixth {
-  width: 100% / 6;
+  width: calc(100% / 6);
```
MichaelWalker: Why do we need the `calc`?
I'd have expected this to be less reliable than the previous version.

directory_components/export_elements/sass/components/boxes/_base.scss
```diff
@@ -29,9 +29,33 @@
...
+
+.message-box-with-icon {
+  @extend .message-box;
+  display: table;
+}
+
+.message-box-icon {
+  display: table-cell;
+  font-family: system-ui;
+  width: 80px;
```
MichaelWalker: I _think_ icons are mostly 64px elsewhere (might be wrong about this :) )
Should we attempt to be consistent with the size of our icons going forward, or do they actually need to be different sizes?

directory_components/templates/directory_components/message_box_with_icon.html
```diff
@@ -0,0 +1,9 @@
+<div class="message-box-with-icon border-medium border-{{ border_colour|default:'teal' }} {{ box_class|default:'width-full background-white' }}">
```
MichaelWalker: As a dev using this component, how should I find out what the valid values are for `border_colour`?

eadpearce: oh good point. i can put this in the docs. they match our sass colour var names that we use in utility classes

directory_components/export_elements/sass/components/boxes/_base.scss
eadpearce: sorry i probably should've put screenshots in this! the width isn't for the icon itself as it's a font symbol so the size is controlled instead by the font size

directory_components/export_elements/overrides/govuk_elements/elements/_layout.scss
eadpearce: chrome was complaining about invalid property name :(

MichaelWalker: weird! I would have sworn that was valid but it clearly I'm going insane!

We should probably use `percentage(1/6)` though.
That way the SASS will compile down to nice, easy (non-calc based) CSS, but we still get to keep our easy to understand source.

eadpearce: oh i didn't know about this one! neat. i'll use this instead

directory_components/static/directory_components/export_elements/stylesheets/govuk-elements-styles.css
```diff
@@ -335,10 +335,10 @@ section {
   width: 50%; }
 
 .width-third {
-  width: 100% / 3; }
+  width: 3333.33333%; }
```
MichaelWalker: ahh! 🤔

The `percentage` function expects a fraction rather than something that is already a percentage.
(ie, `percentage(1/3)` rather than `percentage(100/3)`)

eadpearce: oh omg oh no.... i will fix this lol

eadpearce: good thing we git track non-minified CSS!!!

PR #217, opened 2019-06-14
Framework for less effort tagging.

MichaelWalker:
@eadpearce @richtier @agiamas @Kenneth-Siu @jessgilbert 

~Very much a work in progress at the moment, but pushing it here to get input nice and early.~
_Edit: Now ready to review!_

I normally prefer doing things explicitly rather than trying to have a magic file somewhere that sorts everything out. 

However, we've tried that approach for the GA tagging stuff, and its been unpleasant, and its going to continue to be a pain to maintain.

Edie has already done some great work to move the 'pageLevel' stuff into the CMS and make it all happen automatically. Here is my (first) attempt to do something similar for the 'event' stuff.

The general idea is as follows

- there are lots of types of event which we want basically everywhere (such as a link being clicked)
- with no configuration, we should have a sensible guesses at making an event to record these interactions
- it should be possible to over-write any of these sensible guesses with explicit configuration.

Hopefully this should mean that we;
a) cover 90% of cases without really having to do anything on a case-by-case basis.
b) for the other 10%, allow for full control with easy, explicit configuration.
c) can easily get a sensible baseline, and then gradually add explicit config to improve tracking when required.
d) automatically also tag things embedded in content from the CMS

Config done by something like
```HTML
<section data-ga-section"industries">
    <a href="/all-industries">Click Me</a> # This 'JustWorks'
    {% gaData type="contactButton" %}
        <a href="/contact">Contact</> # This has the type over-written by the configuration
    {% end_gaData %}
</section
```

I'll add documentation for this to the demo app if we think its a good idea! :)


directory_components/static/directory_components/js/dit.tagging.js
```diff
@@ -0,0 +1,190 @@
...
+            addTaggingForForms();
+        });
+
+        function addTaggingForLinks() {
+            var leftMouseButton = 1;
+            var middleMouseButton = 2;
+            var enterKeyValue = 13;
+            $('a')
+                .on('mouseup', function(event) {
+                    if (event.which === leftMouseButton || event.which === middleMouseButton) {
```
richtier: event.which is deprecated. over time we will need to replace usage of `event.which` with something else

https://stackoverflow.com/questions/35394937/keyboardevent-keycode-deprecated-what-does-this-mean-in-practice

for now it's fine though

MichaelWalker: thanks - looks like `event.key === "Enter"` is the new version, which is much nicer anyway (and has good enough browser support) so I'll just change it to that

PR #218, opened 2019-06-17
Add handler for 404 and 500

richtier:


demo/views.py
```diff
@@ -80,3 +82,13 @@ def get_context_data(self, *args, **kwargs):
...
+
+
+class Trigger404View(View):
+    def dispatch(self, request):
+        raise Http404()
+
+
+class Trigger500ErrorView(View):
+    def dispatch(self, request):
+        100/0  # triggering a server error
```
sdonk: Raise an exception?

sdonk: HttpResponseServerError() even better?

PR #220, opened 2019-06-18
Make breadcrumbs more flexible

richtier:
image


CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+- Remove support for python 3.5
+- breadcrumbs template tag interface has changed:
+
+#### before
+1. `{% breadcrumbs left_url='/' left_label='Home' right_label='Statistics' %}`
+2. `{% breadcrumbs left_url='/' left_label='Home' middle_url='/middle/' middle_label='middle' right_label='Statistics' %}`
+
+#### after
+1. `{% breadcrumbs 'Statistics' %}<a href="/">Home</a>{% endbreadcrumbs %}`
+2. `{% breadcrumbs 'Statistics' %}<a href="/">Home</a><a href="/middle/">middle</a>{% endbreadcrumbs %}`
```
MichaelWalker: I'm actually not sure how I feel about this one.

I think what makes me uneasy is that there is quite a large difference between the HTML that you write is a little unnatural, and isn't particularly similar to the output.

ie - there's something that feels slightly wrong to me about writing something like this `<a href="/">Home</a><a href="/middle/">middle</a>` and expecting it to produce a list of things with all the correct classes.

Its clever, but seems a bit surprising and magic to me.

I think I prefer the approach that Edie took in great domestic of passing in a list of 'label/url' pairs.

```
{% breadcrumbs 'Statistics' with parents=[{label: 'Home', url: '/' }, {label: 'middle', url: '/middle/' }] %}
```

But I'd be open to being persuaded! 😊

eadpearce: the template that i made is tailored to work with the `tree_based_breadcrumbs` field that comes from the cms and that does accept the format you suggested. unforch you can't use the syntax in your example in the template or it blows up :( the list of dictionaries would have to be set in the view instead which is a bit of a faff. i think if you're familiar with how the breadcrumbs work this new format makes sense, even if it's doing a lot of magic. as long as the magic is consistent haha

directory_components/templatetags/directory_components.py
```diff
@@ -211,11 +206,55 @@ def render(self, context):
...
+        bit = token.split_contents()[1]
+    except IndexError:
+        raise ValueError('Please specify the label of the current page')
+    current = template.Variable(bit).resolve(None)
+    return Breadcrumbs(nodelist, current=current)
+
+
+class Breadcrumbs(template.Node):
+    template = """
+        <nav aria-label="Breadcrumb" class="breadcrumbs">
```
eadpearce: not sure if this `aria-label` on the nav is necessary but probably wouldn't hurt to have it. govuk design system doesn't seem to use it (though they also just use a `div` and not a `nav`) https://design-system.service.gov.uk/components/breadcrumbs/

tests/test_templatetags.py
```diff
@@ -691,3 +688,28 @@ def test_ga360_tracker_can_optionally_tag_ga_value():
         ' </a>' \
         '</div>'
     assert rendered_html == expected_html
+
+
+def test_bradcrumbs():
+    template = Template(
+        '{% load breadcrumbs from directory_components %}'
+        '{% breadcrumbs "Current Page" %}'
+        '<a href="/foo"></a>'
```
eadpearce: shouldn't this error if any of the links are blank?

CHANGELOG.md
richtier: I'll try and be persuasive. Hold my milk I'm going in.


- *Consistency is nice*: this aims to make `breadcrumbs` tag flexible enough to work for all usecases. Currently, there are cases where the `left`, `middle` and `right` do not apply e.g, CMS breadcrumbs, and pages that want four items.with the new apporach we can easily do for loops the template. This means there are still multiple ways to do breacrumbs.
- *Readability counts*: allow other tags like {% url ... %} to be used samelessly. in the current approach and with the `with parents` approach requires doing `{% url ... as the_url %}` first and then passing `the_url` in. that's hard to read. in a few places we need three urls. not a pleasure to maintain. Take the following real case as an example:
````
  {% url 'investment-support-directory:company-contact' company_number=company.number as contact_url %}
  {% breadcrumbs left_url=services_urls.great_international left_label='Great.gov.uk international' middle_url=contact_url middle_label='Contact company' right_label='Message sent' %}
```

with the new approach

```
  {% breadcrumbs 'Message sent' %}
    <a href="{{ services_urls.great_international }}">Great.gov.uk international</a>
    <a href="{% url 'investment-support-directory:company-contact' company_number=company.number %}">Contact company</a>
    {% endbreadcrumbs %} 
```

you're right there is a cost of the dev scratching their head a little - but on balance this change adds one small problem and removed a few larger onces. not a silver bullet but it is progress.

note too that it's not too magic - it's really "just" wrapping the specified links in 


```
        <nav aria-label="Breadcrumb" class="breadcrumbs">
          <ol>
             <!-- the specified links go here -->
          </ol>
        </nav>
```

directory_components/templatetags/directory_components.py
richtier: FWIW, our current breadcrumbs implementation has one - soI included it for feature parity.

tests/test_templatetags.py
richtier: that's a good idea. will add

demo/templates/demo/breadcrumbs.html
```diff
@@ -20,18 +20,57 @@ <h1 class="heading-large">Breadcrumbs</h1>
...
+  <a href="http://www.example.com">home</a>
+  <a href="http://www.example.com/middle/">middle label</a>
+{% endbreadcrumbs %}
+
       </div>
     </div>
     <div class="grid-row">
       <div class="column-full padding-bottom-15">
         <h2 class="heading-medium">Two levels</h2>
         <p>Similar to three levels, but exclude middle_url and middle_label.</p>
```
MichaelWalker: Could you update the description of this bit of code?

demo/templates/demo/breadcrumbs.html
```diff
@@ -20,18 +20,57 @@ <h1 class="heading-large">Breadcrumbs</h1>
...
+
+        {% breadcrumbs 'current page' %}
+          <a href="http://www.example.com">home</a>
+        {% endbreadcrumbs %}
       </div>
     </div>
+    <div class="grid-row">
+      <div class="column-full padding-bottom-15">
+        <h2 class="heading-medium">More levels</h2>
+        <p>Similar to three levels, but exclude middle_url and middle_label.</p>
```
MichaelWalker: Similarly, this description doesn't quite seem right any more.

CHANGELOG.md
MichaelWalker: Yeah that makes sense, and I agree that its nicer than what we had before by the way, just wasn't convinced that it was the best way of doing it. I don't have a better suggestion though! :)

PR #222, opened 2019-06-24
CI-259 image lazy loading

Kenneth-Siu:
Add lazysizes library, which allows lazy loading of images (and more)

🎉


directory_components/templates/directory_components/base.html
```diff
@@ -47,8 +44,13 @@
             </style>
         </noscript>
 
-        {% block head_js %}
+        {% block vendor_js %}
+            <script src="{% static 'directory_components/js/vendor/lazysizes.min.js' %}"></script>
```
richtier: this will be loaded on every page. do we want that?

Kenneth-Siu: We'll want it on most pages, and it'll end up being cached anyway by the browser.

PR #223, opened 2019-06-26
Bug fixes

eadpearce:


directory_components/templates/directory_components/base.html
```diff
@@ -75,6 +75,7 @@
             <script src="{% static 'directory_components/js/dit.utils.js' %}"></script>
             <script src="{% static 'directory_components/js/dit.classes.Dropdown.js' %}"></script>
             <script src="{% static 'directory_components/js/dit.classes.Modal.js' %}"></script>
+            <script src="{% static 'directory_components/js/dit.components.greatDomesticHeader.js' %}"></script>
```
MichaelWalker: Do we need this here?
I'm guessing that the international apps don't actually want this file, and if so it would be nicer if they didn't have to download this file anyway.

setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='directory_components',
-    version='19.0.0',
+    version='19.0.1',
```
MichaelWalker: Could you add a changelog for this?

directory_components/templates/directory_components/base.html
eadpearce: I assumed this was erroneously removed and it was the header tagging JS that was meant to be removed. If not, this should have been marked as a breaking change in the last version

MichaelWalker: ah - yeah that seems likely! (sorry - sounds like my mistake) :) 

I don't think that this file should be here though (for the reasons above), so perhaps we need a new PR that removes this (marked as a breaking change)

eadpearce: yeah probs. in the past we had more services using the domestic header than the international one so it made sense to include domestic by default and have international services override it. numbers are balancing out now though so yeah i think this makes more sense. i can leave it out in this PR and retroactively update the changelog to add it to breaking changes

MichaelWalker: 🤷‍♂️ I don't feel too strongly about it! Happy for this to be
- left for now
- listed as a breaking change in 19
- a new PR and a breaking change in 20

I'll leave it up to you! 😊

PR #224, opened 2019-06-26
no ticket - Allow compatability with more recent versions of director…

MichaelWalker:
…y components

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import setup, find_packages
 
 setup(
     name='directory_components',
-    version='19.0.1',
+    version='20.0.0',
     url='https://github.com/uktrade/directory-components',
     license='MIT',
     author='Department for International Trade',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='directory_components',
-    version='19.0.1',
+    version='19.0.2',
```
ababic: @MichaelWalker some apps (e.g. `great-domestic-ui`) don't specify a version of `directory-constants`, so upgrading to this version of components could potentially cause breaking changes, because it would also update the `directory-constants` version to a backwards-incompatible version. So, I think this might be better set to version `20.0.0`

richtier: agreed. 20.0.0 :+1:

MichaelWalker: haha - just posted a related comment on slack!

> don't specify a version of directory-constants, so upgrading to this version of components could potentially cause breaking changes

Do you know why do we choose to do it that way?
In general I'd be much happier if apps declared their dependencies (to avoid exactly this kind of thing)

It seems to me that we should want to be able to treat dependency versions as a implementation detail, otherwise it becomes very difficult to reason about the impact of changes

MichaelWalker: (ie, a dev can't make a change here without a fairly deep knowledge of how all the other apps work)

ababic: @MichaelWalker I think I'm in agreement with you now... If apps like `great-domestic-ui` are dependent on a specific version of components, they should specify as much. I think the `19.0.2` version works here.

PR #225, opened 2019-06-28
Make component templates extendable

eadpearce:
allows services to extend the base component templates if they need to modify them slightly without having to rewrite the whole template

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 26.1.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/225/files)
+
+### Implemented enhancements
+- Add overridable blocks to component templates to allow some services to modify them when needed
+
 ## 26.0.0
 [Full Changelog](https://github.com/uktrade/directory-components/pull/241/files)
 
@@ -15,7 +21,7 @@
 [Full Changelog](https://github.com/uktrade/directory-components/pull/240/file)
 
 ## Bugs fixed
-- Update lodash vulnerability 
+- Update lodash vulnerability
 
 ## 25.0.0
 [Full Changelog](https://github.com/uktrade/directory-components/pull/228/files)
--- a/directory_components/templates/directory_components/card.html
+++ b/directory_components/templates/directory_components/card.html
@@ -10,12 +10,15 @@
     href="{{ url }}"
     {% if card_id %}id="{{ card_id|add:'-link' }}"{% endif %}>
   {% endif %}
+    {% block image %}
     {% if img_src %}
       {% lazyload %}
         <img class="card-image" src="{{ img_src }}" alt="{{ img_alt }}" {% if card_id %}id="{{ card_id|add:'-image' }}"{% endif %}>
       {% endlazyload %}
     {% endif %}
+    {% endblock %}
     <div class="card-inner">
+      {% block content %}
       {% if html_content %}
       {{ html_content }}
       {% else %}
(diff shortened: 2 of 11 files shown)


directory_components/templates/directory_components/card.html
```diff
@@ -7,10 +7,13 @@
     href="{{ url }}"
     {% if card_id %}id="{{ card_id|add:'-link' }}"{% endif %}>
   {% endif %}
+    {% block image %}
     {% if img_src %}
       <img class="card-image" src="{{ img_src }}" alt="{{ img_alt }}" {% if card_id %}id="{{ card_id|add:'-image' }}"{% endif %}>
     {% endif %}
+    {% endblock %}
     <div class="card-inner">
+      {% block content %}
```
MichaelWalker: I think we already have a `content` block on `base.html`.

Could we use a different name here to avoid a clash?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 20.1.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/225/files)
+
+### Implemented enhancements
+- Add overridable blocks to component templates to allow some services to modify them when needed
```
MichaelWalker: @jessgilbert and I were talking about this exact thing yesterday!

We were thinking along very similar lines, but had a slightly different implementation (although we haven't actually written any code for it)

If my understanding of Django templates is correct, the code below will break if there is more than 1 card on the page, as you can't have multiple blocks with the same name. 

Is that right, or have I missunderstood something?

MichaelWalker: Instead, Jess and I were suggesting something like...

- Defining a `base_card` django tag, which  would take whatever was inside it, and injecting into a some HTML for the card + image (+ title??)
- perhaps adding some logic to apply sensible css classes to the contents of the cards
- re-writing the existing card templates to use the above

That way, almost all of the time we could continue to use
``` {% card url=foo etc %} ```

But we'd also be able to handle more unusual cards using the exact same code like
```
{% base_card url=foo etc %}
    <div>Lots of content goes here</div>
{% end_base_card %}
```

Our hope was that we'd be able to collapse all of our card variants down to a couple of templates plus this flexible base class. 

And we'd therefore be better able to keep everything up to date.

Does that make any sense?

MichaelWalker: Base Tag would look a little like the existing breadcrumbs one

eadpearce: Nope since block names would only clash if they're actually written in the same template. Since cards are only used as includes this won't happen - it's like variable scope. The card's `content` block also won't clash with `base.html`'s `content` block

directory_components/templates/directory_components/card.html
eadpearce: It won't clash but we could call it something like `component_content` or `content_override` if you like

CHANGELOG.md
eadpearce: I'm not sure I follow the implementation but it sounds like a useful thing to have! Does that mean we can use the modified templates with the same kind of inclusion tags?

MichaelWalker: Awesome. I didn't realise that templates were scoped like that, but it makes total sense! Thanks :) 

Yep I think my ultimate goal was to make it easy for all of the card variants in all of the projects to use as much common stuff as possible, while hopefully also reducing the number of files with names like `card_with_this_special_little_thing.html`.

(I started writing some code here, but perhaps easiest if I just do a little spike and throw up a PR so that we can see if it works!)

PR #232, opened 2019-07-04
Created new search page components section including selected filters…

jessgilbert:
… and expandable options
screencapture-localhost-9013-search-page-components-2019-07-04-16_03_06
image

 - [x] Changelog entry added.


demo/templates/demo/search-page-components.html
```diff
@@ -0,0 +1,77 @@
...
+
+{% load static %}
+{% load demo_tags %}
+{% load breadcrumbs search_page_selected_filters search_page_expandable_options from directory_components %}
+
+{% block css_layout_class %}components{% endblock css_layout_class %}
+
+{% block body_header %}
+  {{ block.super }}
+  {% include 'directory_components/header_beta_banner.html' %}
```
MichaelWalker: probably just don't need this beta banner.

demo/templates/demo/search-page-components.html
```diff
@@ -0,0 +1,77 @@
...
+{% endblock %}
+
+{% block content %}
+
+<div class="container">
+  {% breadcrumbs 'Search page components' %}<a href="/">Home</a>{% endbreadcrumbs %}
+</div>
+
+<div class="container">
+    <h1 class="heading-xlarge">Search page components</h1>
```
MichaelWalker: Could we add a description of what the `filters` object should be?

directory_components/templates/directory_components/search-page-selected-filters.html
```diff
@@ -0,0 +1,11 @@
+{% if filters %}
+    <div class="selected-filters-container underline-flag-red margin-bottom-30">
+        <div class="padding-15">
```
MichaelWalker: I'd be tempted to hard code more of these things.

- `selected_filterers_title` and `clear_all_link_text` can probably be hard coded here - we always want to them to be the same, and it would allow us to provide the translations for them here rather than having to get the translations in all of the apps that include this component.
- I _think_ `clear_all_link` will always just be `/`, so we could hard code it to that - if it ever needs to be anything else we can make this component less restrictive later.

In general I think we should be trying to lock components down so they have as few moving parts as possible initially, and then opening them up if/when it becomes useful.

directory_components/templates/directory_components/search-page-expandable-options.html
```diff
@@ -0,0 +1,16 @@
+<fieldset class="expandable-fieldset">
+    <div class="">
```
MichaelWalker: empty class looks a little odd.

demo/templates/demo/search-page-components.html
```diff
@@ -0,0 +1,77 @@
...
+        </div>
+    </div>
+
+    <h2 class="heading-medium">Expandable options</h2>
+
+    <pre><code>{% verbatim %}{% load search_page_expandable_options from directory_components %}
+
+&lt;div class="grid-row"&gt;
+    &lt;div class="column-quarter"&gt;
+        {% search_page_expandable_options class_width="column-quarter" option_title="Sector" options=options %}
```
MichaelWalker: I don't think we use the `class_width` at the moment.

demo/templates/demo/search-page-components.html
```diff
@@ -0,0 +1,77 @@
...
+    &lt;/div&gt;
+&lt;/div&gt;{% endverbatim %}</code></pre>
+
+
+    <div class="grid-row">
+        <div class="column-quarter">
+            {% search_page_expandable_options class_width="column-quarter" option_title="Sector" options=options %}
```
MichaelWalker: Could we add a bit which explains what the options should be?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 20.2.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/232/files)
+### Implemented enhancements
+- Added new `Search page components` including the `selected filters` and `expandable options` components
```
MichaelWalker: These look awesome - I think there are even more things that we could add though (in this PR or another one)

This would be good to have:
image

This bit _might_ be good, but is currently not quite consistent across the components - perhaps we should move it here to force it to be consistent. 
image

jessgilbert:
screencapture-localhost-9013-search-page-components-2019-07-05-16_33_57
@MichaelWalker

directory_components/templates/directory_components/search-page-selected-filters.html
richtier: if we set `clear_all_link` hard coded to `/` then that would take the user to great domestic home page e.g, if the user was on https://www.great.gov.uk/trade/ or https://www.great.gov.uk/investment-support-directory/ then they would be sent to https://www.great.gov.uk/

jessgilbert: Yes agreed, I have updated since this and am keeping the link to be a variable you need to add in, but have hard coded the text and title👍

PR #233, opened 2019-07-05
Pagination

richtier:
- [x] Changelog entry added.
 - [x] CSS have been compiled.
 - [x] Change is accessible to the standards we subscribe to.

image


directory_components/templates/directory_components/pagination.html
```diff
@@ -0,0 +1,51 @@
...
+    PREV 1 2 3 ... 5 NEXT
+
+    if there are lots of items then show multiple ellipsis:
+         1 ... 8 9 10 ... 15      NEXT
+    PREV 1 ... 8 9 10 ... 15      NEXT
+    PREV 1 ... 8 9 10 ... 15 
+{% endcomment %}
+
+{% if pagination.paginator.num_pages > 1 %}
+    <div class="pagination"">
```
MichaelWalker: Could this be a `nav` instead of a `div`?
(Not sure if that's a thing for pagination, but feels to me like it ought to be)

directory_components/templates/directory_components/pagination.html
```diff
@@ -0,0 +1,51 @@
...
+        {% if pagination.has_previous %}
+            <a class="button-ghost-blue pagination-previous" href="{{ url }}&{{ page_param_name }}={{ pagination.previous_page_number}}">Previous</a>
+        {% endif %}
+            <ol class="{{ pagination.has_previous|yesno:'has-previous ,' }}{{ pagination.has_next|yesno:'has-next,' }}">
+                {% if pagination.paginator.num_pages <= 5 %}
+                    {% comment %} 1 2 3 4 5 {% endcomment %}
+                    {% for number in pagination.paginator.page_range %}
+                        {% include 'directory_components/pagination-number.html' %}
+                    {% endfor %}
+                {% elif pagination.number < 5 %}
```
MichaelWalker: I think if there are 4 pages this is going to give
`1 2 3 4 ... 4`
similar issue for 5 (and arguably 6)?

MichaelWalker: Might be slightly odd if there are 6 pages
(ie `1 2 3 4 ... 6` might as well be `1 2 3 4 5 6`)

directory_components/templatetags/directory_components.py
```diff
@@ -192,6 +193,24 @@ def case_study(**kwargs):
     return kwargs
 
 
+@register.inclusion_tag(
+    'directory_components/pagination.html', takes_context=True
+)
+def pagination(context, objects, current_page, page_param_name='page'):
```
MichaelWalker: is there a reason to pass in the list of objects here rather than just `number_of_pages`?
It seems like we don't really care what the options are, but at the moment we are typing `range` quite a lot to make this work.

directory_components/templates/directory_components/pagination-number.html
```diff
@@ -0,0 +1,9 @@
+{% if ellipsis_before %}
+    <li class="active-blue-text margin-right-0 margin-top-15 margin-15-m"><span>...</span></li>
+{% endif %}
+
+<li><a href="{{ url }}&page={{ number }}" class="{% if pagination.number == number %}button{% else %}button-ghost-blue{% endif %}">{{ number }}</a></li>
+
+{% if ellipsis_after %}
+    <li class="active-blue-text margin-right-0 margin-top-15 margin-15-m"><span>...</span></li>
+{% endif %}
```
MichaelWalker: Would it be better to pull out an `pagination-ellipsis` template too?

That way,
- the page number becomes simpler
- we get rid of the duplication
- the pagination becomes a little more obvious, with

```
{% include number %}
{% include ellipsis %}
{% include number %}
```

MichaelWalker:
Not a huge deal, but I think I would have expected the numbers to always remain in the middle of the page

ie 
```
[N]    [1] [2] [3]
[N]    [1] [2] [3]    [P]
       [1] [2] [3]    [P]
```

rather than
```
[N]        [1] [2] [3]
[N]    [1] [2] [3]    [P]
    [1] [2] [3]       [P]
```

Very happy to be outvoted on that though!

demo/templates/demo/pagination.html
```diff
@@ -0,0 +1,58 @@
...
+
+{% block content %}
+<div class="container">
+  {% breadcrumbs 'Pagination' %}<a href="/">Home</a>{% endbreadcrumbs %}
+</div>
+<section class="padding-top-30">
+  <div class="container">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+        <h1 class="heading-xlarge">Pagination</h1>
```
MichaelWalker: Could we add a little bit of information about how to use the component?

directory_components/templates/directory_components/pagination.html
richtier: yep, that seems apt https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

directory_components/templates/directory_components/pagination.html
richtier: following the requirements set by the designer :)

richtier: but I see what you mean. will run it by him

directory_components/templatetags/directory_components.py
richtier: I see what you mean - we can pass in `number_of_total_results` instead of `objects` (note the Paginator works out number_of_pages for us)

directory_components/templates/directory_components/pagination-number.html
richtier: while that would be DRYer, it would be more complex:

DRY approach:
 - separate ellipsis.html file
 - creation of ellipsis templatetag/directory_components.py
 - loading and using template tag in the number.html template
 - testing template tag
 
...WET... approach:
 - small amount of localised easily maintained duplication


given that this duplication of HTML is all in the same file within a few lines of each other I think on balance small amount of duplication reduces overall maintenance cost:

richtier:
>  but I think I would have expected the numbers to always remain in the middle of the page

the design expected the entire paginator to be centred. that looks odd in the demo because there are multiple paginators. In practice, though there will only be one per page.

For nice demo though I've added style to make the numbers remain in the middle.

PR #235, opened 2019-07-09
New International Header

MichaelWalker:
We need a new header on international to cope with an upcoming IA changes.
This PR creates a new 2 level navigation for the international site (see below)

For now though, the navigation will only use the first level of the nav, so that initially it will be only a very minor change from the current appearance, although the code will allow us to easy update and extend going forwards

We've also commonised the domestic and international headers a little more, and improved the non-js behaviour of both headers for mobile.

Credit for all the hard work here goes to @Kenneth-Siu 

 - [X] Changelog entry added.
 - [X] (if adding CSS) CSS have been compiled.
 - [X] (if adding HTML) Change is accessible to the standards we subscribe to.
 - [X] (if changing the UI) Add a printscreen to the PR


MichaelWalker:
GifMaker_20190709100128704

MichaelWalker:
GifMaker_20190709100128704
GifMaker_20190709100209518

directory_components/mixins.py
```diff
@@ -121,3 +124,70 @@ def get_context_data(self, *args, **kwargs):
...
+            HeaderItem(
+                title=page.title,
+                url=page.url,
+                active=page.name == self.header_section)
+            for page in self.pages
+        ]
+
+        active_page = next(
+            (page for page in self.pages if page.name == self.header_section),
+            None)
```
richtier: bit hard to read. a function would be easier to read. would also allow overriding the behaviour, which may be useful?
```
def get_active_page(self):
    for page in self.pages:
        if page.name == self.header_section:
            return page
```

directory_components/mixins.py
```diff
@@ -121,3 +124,70 @@ def get_context_data(self, *args, **kwargs):
             *args,
             **kwargs
         )
+
+
+Page = namedtuple('Page', 'name title url sub_pages')
+HeaderItem = namedtuple('HeaderItem', 'title url active')
```
richtier: `is_active` rather than `active` would help the reader better understand this is a boolean

directory_components/static/directory_components/js/dit.components.header.js
```diff
@@ -0,0 +1,26 @@
...
+dit.components = dit.components || {};
+dit.constants = dit.constants || {};
+
+dit.components.header = (new function() {
+  const self = this;
+
+  self.init = function() {
+    const mobileMenuButtons = document.getElementsByClassName("mobile-menu-button");
+    for (let i = 0; i < mobileMenuButtons.length; i++) {
+      const mobileMenuButton = mobileMenuButtons[i];
```
richtier: please use `var` - many of the browsers we support do not have `const` yet :)

MichaelWalker: good spot. thanks

PR #236, opened 2019-07-09
Updated expandable options to use form

jessgilbert:
screencapture-localhost-9013-search-page-components-2019-07-09-11_11_44

 - [x] Changelog entry added.


demo/views.py
```diff
@@ -46,12 +46,7 @@ class SearchPageComponentsDemoPageView(BasePageView):
     def get_context_data(self, *args, **kwargs):
         context = super().get_context_data(
             filters=['Energy', 'Real Estate', 'Automotive', 'Aerospace'],
-            options={
-                'Energy': 'checked',
-                'Real Estate': '',
-                'Automotive': '',
-                'Aerospace': 'checked'
-            },
+            form=forms.MultipleChoiceForm,
```
richtier: in practice, the forms passed to template would be instantiated, so please do `forms.MultipleChoiceForm()` :)

jessgilbert:
@richtier Have changed the changelog

PR #238, opened 2019-07-09
Made a javascript file that needs to be loaded on the page when expan…

jessgilbert:
…dable filters component is used

screencapture-localhost-9013-search-page-components-2019-07-09-14_32_31

 - [x] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,7 @@
 - Moved javascript from actual component `search-page-expandable-options.html`to `dit.components.searchPageExpandableOptions.js` which needs to be loaded when component is used (fixes bug when component is used more than once on a page)
 
 ### Breaking changes
-- Requires javascript file to be loaded when component is used now
-
+- Requires javascript file `dit.components.searchPageExpandableOptions.js` to be loaded when `search-page-expandable-options` component is used
 ## 21.0.1
 [Full Changelog](https://github.com/uktrade/directory-components/pull/237/files)
 ### Bug fixes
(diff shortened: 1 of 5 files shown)


directory_components/static/directory_components/js/dit.components.searchPageExpandableOptions.js
```diff
@@ -0,0 +1,21 @@
+
+dit.components.searchPageExpandableOptions = new function(){
```
MichaelWalker: I _think_ that this is the same, but with less code.
What you've done looks great (and doesn't require jquery, which is nice), but just to show that it's possible to condense this if you wanted to!

```javascript
$(document).ready(
    $('.filter-collapse').on('click', function() {
        $(this).toggleClass('checked');
        var formGroup = document.getElementById($(this).attr('aria-controls'));
        formGroup.setAttribute('aria-expanded', !(formGroup.getAttribute('aria-expanded') == 'true'));
    });
);
```

Up to you what you prefer! :)

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 22.0.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/238/files)
+### Bug fixes
+- Moved javascript from actual component `search-page-expandable-options.html`to `dit.components.searchPageExpandableOptions.js` which needs to be loaded when component is used (fixes bug when component is used more than once on a page)
+
+### Breaking changes
+- Requires javascript file to be loaded when component is used now
```
MichaelWalker: Can we make this a little more descriptive (it makes perfect sense with the lines above, but I suspect lots of people might only read the breaking changes and then get confused! 😂 )

PR #239, opened 2019-07-10
Pagination component takes in Pagination object rather than creating …

jessgilbert:
…its own

 - [x] Changelog entry added.

Desktop:
screencapture-localhost-9013-pagination-2019-07-10-13_59_47

Mobile:
image
image
image


demo/config/urls.py
```diff
@@ -133,7 +133,7 @@
     ),
     url(
         r'^pagination/$',
-        views.BasePageView.as_view(),
+        views.DemoPaginationView.as_view(),
         {'template_name': 'demo/pagination.html'},
```
richtier: can you move `template_name = 'demo/pagination.html'` to the view please.

we specify the template in urls.py of `BasePageView` urls because we're naughty and it's convenient when using `BasePageView`, but now we've subclassed a TemplateView, we can specify the template_name on the view :)

jessgilbert: Woops I already have in the view, will delete it from here 👍

demo/views.py
```diff
@@ -120,3 +121,40 @@ def dispatch(self, request):
...
+
+
+class DemoPaginationView(TemplateView):
+    template_name = 'demo/pagination.html'
+
+    objects = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
+
+    @property
+    def paginator_few(self):
+        return Paginator(self.objects, 10)
```
richtier: `pagination.html` assumes a page size of 5 (the logic is hard-coded in the template) - so having a different page size in the demo could mislead devs a bit

this could be `Paginator(range(25), 5)`, `Paginator(range(40), 5)`, `Paginator(range(60), 5)`

also due to this page size constraint we should also check in the pagination template tag that the paginator's page size is indeed 5 :)

tests/test_templatetags.py
```diff
@@ -809,13 +811,19 @@ def test_ga360_data_with_all_optional_parameters():
 def test_pagination(count, current, expected, rf):
     template = Template(
         '{% load pagination from directory_components %}'
-        '{% pagination objects_count=objects_count current_page=current %}'
+        '{% pagination pagination_page=pagination_page %}'
     )
 
+    page_size = 5
+    objects = []
+    for i in range(count):
```
MichaelWalker: there are a few shorter options here if you want them

```python
objects = [""] * count
objects = ["" for _ in range(count)]
objects = [item for item in range(count)]
```

demo/views.py
```diff
@@ -120,3 +121,40 @@ def dispatch(self, request):
...
+    def paginator_many(self):
+        return Paginator(self.objects, 1)
+
+    @property
+    def pagination_few_pages(self):
+        return [self.paginator_few.page(index) for index in range(1, 3, 1)]
+
+    @property
+    def pagination_some_pages(self):
+        return [self.paginator_some.page(index) for index in range(1, 6, 1)]
```
richtier: it's a bit hard to read with the `paginator_some` being defined as a seperate property. this would be easier:
```
    @property
    def pagination_some_pages(self):
        paginator = return Paginator(self.objects, 3) 
        return [paginator.page(index) for index in range(1, 6, 1)]
```

richtier: also range:
```
range([start], stop[, step])
```
`step` is 1 by default so we can do `range(1, 6)`

demo/views.py
jessgilbert: The page size is no longer hard coded in the inclusion tag, the page size is included in the pagination so a page can choose its own page size

PR #241, opened 2019-07-11
Simplify forms and add pagination helper

richtier:
- [x] Changelog entry added.

image


directory_components/forms/__init__.py
```diff
@@ -0,0 +1,3 @@
+from directory_components.forms.fields import * # NOQA
+from directory_components.forms.forms import * # NOQA
+from directory_components.forms.widgets import * # NOQA
```
richtier: this will allow us to interact with our forms in the same way we would interact with [django](https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/forms/__init__.py)

so we can now do 
```
from directory_component import forms

class MyForm(forms.Form):
    field = forms.CharField(
        widget=forms.PrettyIDsMixin()
    )
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+## 26.0.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/230/files)
```
jessgilbert: Wrong pr number :)

richtier: oy vey :) thanks

directory_components/helpers.py
```diff
@@ -222,3 +222,13 @@ def values_to_labels(values, choices):
...
+
+
+def get_pagination_url(request, page_param_name):
+    """Remove pagination param from request url"""
+    url = request.path
+    params = request.GET.copy()
+    params.pop(page_param_name, None)
+    if params:
+        url = f'{url}?{params.urlencode()}'
+    return url
```
jessgilbert: I thought this method was already part of directory-components? I haven't noticed if this has been moved from another file or was this not already in directory-components?

richtier: it was previously within the `pagination` template tag - bug now this code can be used by `pagination`  and anywhere else (great-international and FAS/ISD will find it useful)

jessgilbert: Ah I see now 👍 All looks sensible to me 😄

PR #242, opened 2019-07-16
Display package version on demo homepage

eadpearce:


demo/views.py
```diff
@@ -18,6 +19,14 @@ def template_name(self):
         return self.kwargs.get('template_name')
 
 
+class IndexPageView(BasePageView):
+
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        context['version'] = get_distribution('directory_components').version
+        return context
```
richtier: this is fine :). FWIW, this would be more *djangoistic*  
```
    version = get_distribution('directory_components').version
    return super().get_context_data(version=version*args, **kwargs)
```

PR #247, opened 2019-07-25
Refactor Domestic Header CSS

MichaelWalker:

 - [x] Changelog entry added.


directory_components/export_elements/sass/components/header-footer/_domestic-header.scss
```diff
@@ -3,45 +3,13 @@
...
-      &:active:focus span {
-        color: $great-blue;
-      }
-    }
-
-    .account-link {
-      color: $stone;
-      margin-left: 5px;
-    }
+#great-domestic-header {
```
eadpearce: please don't use IDs for selectors!

directory_components/export_elements/sass/components/header-footer/_international-header.scss
```diff
@@ -3,8 +3,8 @@
 @import 'header-logo';
 @import 'mobile-menu-button';
 
-.great-header.international {
-  .js-menuitem {
+#great-international-header {
```
eadpearce: and again here

directory_components/templates/directory_components/header_footer/domestic_header.html
```diff
@@ -1,21 +1,18 @@
 {% load static %}
 
-<header id="great-header" class="great-header domestic" dir="ltr" data-ga-section="header">
+<header id="great-domestic-header" class="great-header" dir="ltr" data-ga-section="header">
   {% include 'directory_components/header_footer/global_header.html' with header_class="domestic" %}
 
-  <div id="great-header-menu" class="great-header-menu domestic">
+  <div class="menu">
```
eadpearce: can we keep these classes namespaced under `great-header`? `menu` may be too generic a classname if we want to avoid style pollution

directory_components/templates/directory_components/header_footer/domestic_header.html
```diff
@@ -1,21 +1,18 @@
...
         {% static 'directory_components/images/eig-logo-stacked.svg' as logo_image_url %}
         {% include 'directory_components/header_footer/header_logo.html' with url=service_urls.exred image_source=logo_image_url image_alt_text='Exporting is Great' %}
       {% endblock %}
 
-      <div class="great-header-extra-links" id="great-header-extra-links">
-        <ul>
-          {% include 'directory_components/header_footer/sso_login.html' %}
-        </ul>
-      </div>
+      <ul class="extra-links">
```
eadpearce: same here

directory_components/export_elements/sass/components/header-footer/_domestic-header.scss
MichaelWalker: Out of interest, what't your reasoning for this?

I did this very deliberately. We have some styling that we apply to all 'great-header' classes, which is applied based on the class name.

We then have a single specific version of that header that is tagged and styled using its ID.
Doing so means that
- styling for that object will be scoped to that specific object. There's no chance of that styling leaking out into anywhere else
- styling for that object will superseed styling in the 'shared great-header' styling (and this will be robust - without it, adding another class in the shared style could change the CSS precedence and break things without warning)

I agree that using class names for styling in general gives you better flexibility, but to me this particular use case feel like a fair use of ID selectors.

directory_components/templates/directory_components/header_footer/domestic_header.html
MichaelWalker: all the new styles are scoped to only ever apply in the header, so they shouldn't pollute other parts of the code base.

I guess I'm relying on other things not polluting this though, which isn't perfect.
I'll can add back some great-header's when I get some time?

Depends what SCSS patterns we like I guess! :)

PR #254, opened 2019-08-01
Added settings cleanup management commands

richtier:
- [x] Changelog entry added.

### settings_shake
image

## environment_diff
image


MichaelWalker:
What's the reason for running this as a management command?

In my head, this is a thing that 
a) is totally standalone and has nothing to do with directory components
b) is a thing we'd only ever want to run locally (as it has nothing to do with the server its running on and requires our personal tokens)

Would it not be better for this to be just a plain script that doesn't need the hassle of Django?

Also I'd slightly rather not introduce this to directory components. It doesn't feel like its particularly related to anything here, and it looks like the current setup would essentially cause there to the additional dependencies for all the deployed apps that use directory components (as hvac and vulture will now be required everywhere)

MichaelWalker:
Wasn't there some work happening to get the diff over time from Vault? Would this be better off living with that?

directory_components/management/commands/environment_diff.py
```diff
@@ -0,0 +1,68 @@
...
+    help = 'Diff the vault of two environments.'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--token',
+            help='Vault token. Retrieve by clicking "copy token" on Vault UI.'
+        )
+        parser.add_argument(
+            '--domain',
+            default=getattr(settings, 'DIRECTORY_COMPONENTS_VAULT_URL', None),
```
MichaelWalker: Is there actually any advantage to not just hard coding this?
Vault's URL isn't secret, and isn't likely to change.

Even if it did change, we'd want it for all apps, so it would be easier to change the hard-coded value.

(I'm still fine with having the command line argument in case you wanted to manually overwrite it though :) )

MichaelWalker: similar argument for the Root Path?

directory_components/management/commands/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+
+def prompt_user_choice(message, options):
+    display = '\n'.join([f'{[i]} {option}' for i, option in enumerate(options)])
+    index = int(input(f'{message}:\n\n{display}\n\n'))
+    return options[index]
+
+
+def clean_secrets(secrets):
```
MichaelWalker: What are we hoping to protect against here?
I'd argue that;

- its only being used locally, (by people that have a token, and therefore have access to see all the things anyway), so there isn't much need to filter out the secret secrets.
- it is useful to be able to see if the secret secrets differ, so would be nice to have them
- if we are concerned about security, then this probably isn't reliable enough to meet our need.

directory_components/management/commands/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    ignore_settings = getattr(
+        settings,
+        'DIRECTORY_COMPONENTS_VAULT_IGNORE_SETTINGS_REGEX',
+        DEFAULT_UNSAFE_SETTINGS
+    )
+    for key in secrets.copy():
+        for entry in ignore_settings:
+            if entry.match(key):
+                del secrets[key]
```
MichaelWalker: also, if we decide we do want this, could we instead only delete the value (or replace with stub value), rather than the entire entry.

That way we'd still be able to see if its present on one environment but missing in another.

directory_components/management/commands/helpers.py
```diff
@@ -0,0 +1,76 @@
...
+    return secrets
+
+
+def get_secrets(client, path):
+    response = client.read(path=path)
+    return clean_secrets(response['data'])
+
+
+def diff_dicts(dict_a, dict_b):
+    differ = difflib.Differ()
```
MichaelWalker: we never seem to use the `differ`

directory_components/management/commands/settings_shake.py
```diff
@@ -0,0 +1,117 @@
...
+        self.stdout.write(label)
+
+        if options['wizard']:
+            secrets = helpers.get_secrets_wizard(client=client, root=options['root'])
+        else:
+            secrets = helpers.get_secrets(
+                client=client,
+                path=f"{options['root']}/{options['project']}/{options['environment']}",
+            )
+        return [key for key in secrets if not hasattr(settings, key)]
```
MichaelWalker: Nice! :)

directory_components/management/commands/environment_diff.py
richtier: > Vault's URL isn't secret, and isn't likely to change.
you guessed my answer. I'm treating it as a secret. I'll confirm with webopps

directory_components/management/commands/helpers.py
richtier: - secrets are _should_ be different in different environments so the reader gains no info seeing them. it's filtering out noise.
- cheeky print screens and hasty copy-and-paste would be disastrous for leaking secrets.

directory_components/management/commands/helpers.py
richtier: stub is useful. will do :+1:

directory_components/management/commands/helpers.py
richtier: good catch

directory_components/management/commands/environment_diff.py
richtier: the advice I got was
> vault is on private VPC behind VPN and cannot be access from Internet, but I wouldn’t put the url in any places are publicly available

I agree: bad actor can compile list of dozens of tiny pieces of information to perform at the very least social engineering. "Hey it's Dave from x can I have Y for <pseud secret here>"

directory_components/management/commands/helpers.py
MichaelWalker: Yep, the cheeky print screen argument is a good one.

directory_components/management/commands/environment_diff.py
MichaelWalker: Yeah, I forgot that it wasn't a publicly hosted service so all of this makes sense to me too! 👍

I double checked the organisation, and we don't currently expose it in any of our public repos, so lets keep it that way! 😊

PR #259, opened 2019-08-13
Reduce false positives of redundant vault values

richtier:
image

### DIFF
--- a/directory_components/janitor/management/commands/settings_shake.py
+++ b/directory_components/janitor/management/commands/settings_shake.py
@@ -1,12 +1,10 @@
 import hvac
 
 from django.conf import global_settings, settings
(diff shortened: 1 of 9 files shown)


directory_components/janitor/management/commands/settings_shake.py
```diff
@@ -78,9 +78,12 @@ def report_obsolete_vault_entries(self, options):
                 client=client,
                 path=f"{options['root']}/{options['project']}/{options['environment']}",
             )
-        return [key for key in secrets if not hasattr(settings, key)]
+        # there is no guarantee env vars and the settings variable name jave the same name,
```
eadpearce: "have"?

PR #260, opened 2019-08-14
GTRANSFORM-346 update card component to allow external links, add ext…

jeanettemariemurphy:
…ernal links to footer

 - [x] Changelog entry added.


directory_components/templates/directory_components/labelled_image_card.html
```diff
@@ -2,6 +2,7 @@
   class="labelled-image-card"
   href="{{ url }}"
   {% if card_id %}id="{{ card_id|add:'-link' }}"{% endif %}
+  {% if link_type %}target="_blank" rel="noopener noreferrer" title="Opens in a new window"{% endif %}
```
eadpearce: ah. since you're only checking for the truthyness of this flag it might make more sense for it to be passed to the template as `external_link=True` and otherwise it defaults to false? since it's only ever going to be on or off anyway a bool makes sense

demo/templates/demo/cards.html
```diff
@@ -127,6 +127,14 @@ <h2 class="heading-medium">Labelled linked card with title (if no image is set)<
       </div>
     </div>
 
+    <div class="grid-row">
+      <div class="column-half">
+        <h2 class="heading-medium">Linked card with external link, opens in new window</h2>
+        {% card card_id='card-example-image-description' img_src=card_image02 url='#' description='Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.' external_link='True' %}
+        <pre><code>{% verbatim %}{% card card_id="card-example-image-description" img_src=card_image02 url='#' description='Lorem ipsum dolor sit amet...' external_link='True' %}{% endverbatim %}</code></pre>
```
eadpearce: this should be without quotes `external_link=True`

tests/test_templatetags.py
```diff
@@ -346,6 +346,29 @@ def test_card_with_icon():
...
 
 
+def test_card_with_external_link():
+    card_content = {
+        'external_link': 'True',
+        'url': 'url',
+    }
+    string = (
+        "{{% load card from directory_components %}}"
+        "{{% card external_link='{external_link}' url='{url}' %}}"
```
eadpearce: and same here `external_link={external_link}`

PR #263, opened 2019-08-16
Persist other query params on lang selector submit

eadpearce:

 - [x] Changelog entry added.


directory_components/templates/directory_components/header_footer/language_switcher.html
```diff
@@ -4,6 +4,11 @@
   <div id="great-header-language-switcher" class="great-header-language-switcher great-header-form">
     {% language request.LANGUAGE_CODE %}
       <form class="languages" action="" method="get">
+        {% for name, value in language_switcher.hidden_fields %}
```
richtier: I think we can stop passing `hidden_fields` and this can be 
```
{% for key, value in request.GET.items %}
    <input type="hidden" name="{{ name }}" value="{{ value }}" />
{% endfor %}
```

eadpearce: oh yeah that works. can do

directory_components/templates/directory_components/header_footer/country_switcher.html
```diff
@@ -3,11 +3,16 @@
 {% if features.COUNTRY_SELECTOR_ON and hide_country_selector %}
   <div id="great-header-country-selector" class="great-header-country-selector great-header-form">
     <form class="countries" action="" method="get">
+      {% for key, value in request.GET.items %}
```
kowalcj0: Will this allow ppl to persist an arbitrary number of unverified fields on our form?

eadpearce: nah. it only persists them in the url. this is so we can persist marketing utm codes. the forms still only use the `lang` and `country` params. they ignore but keep everything else

kowalcj0: :+1:

PR #264, opened 2019-08-19
Remove remaining lazyload code from templates

eadpearce:

 - [x] Changelog entry added.

Screen Shot 2019-08-19 at 12 26 08


setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='directory_components',
-    version='27.9.0',
+    version='27.9.2',
```
jessgilbert: How come it goes from 27.9.0 to 27.9.2 in just one update? Should it not be 27.9.1? Even so, is this not more than a minor change as its fixing quite a lot of things?

eadpearce: yeah it was .2 because i was expecting the other PR to be merged first. oops

PR #268, opened 2019-08-21
CI-437: Ensure `django_language` cookie is secure and httponly.

MichaelWalker:

 - [x] Changelog entry added.


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 27.11.1
+[Full Changelog](https://github.com/uktrade/directory-components/pull/266/files)
+
+### Implemented enhancements
+- `django_language` cookie is now set as secure and http-only.
+
 ## 27.10.0
```
MichaelWalker: @jeanettemariemurphy I think we are missing the changelog for 27.11.0.
I think this might have been your change - would you mind adding one?

directory_components/middleware.py
```diff
@@ -127,7 +127,9 @@ def process_response(self, request, response):
             value=translation.get_language(),
             max_age=settings.LANGUAGE_COOKIE_AGE,
             path=settings.LANGUAGE_COOKIE_PATH,
-            domain=settings.LANGUAGE_COOKIE_DOMAIN
+            domain=settings.LANGUAGE_COOKIE_DOMAIN,
+            secure=True,
```
richtier: this would prevent it from working on local dev env as that does not use https. we can do `secure=settings.LANGUAGE_COOKIE_SECURE` though and bob's your uncle

directory_components/middleware.py
```diff
@@ -102,7 +102,7 @@ def process_response(self, request, response):
                 value=request.COUNTRY_CODE,
                 max_age=settings.LANGUAGE_COOKIE_AGE,
                 path=settings.LANGUAGE_COOKIE_PATH,
-                domain=settings.LANGUAGE_COOKIE_DOMAIN
+                domain=settings.LANGUAGE_COOKIE_DOMAIN,
```
richtier: do we want to see `secure=...` here too?

directory_components/middleware.py
```diff
@@ -127,7 +127,9 @@ def process_response(self, request, response):
             value=translation.get_language(),
             max_age=settings.LANGUAGE_COOKIE_AGE,
             path=settings.LANGUAGE_COOKIE_PATH,
-            domain=settings.LANGUAGE_COOKIE_DOMAIN
+            domain=settings.LANGUAGE_COOKIE_DOMAIN,
+            secure=True,
+            httponly=True,
```
richtier: this will prevent javascript from reading the cookie. this ok? @eadpearce

eadpearce: we only read the SSO cookie with JS so this is probably fine

directory_components/middleware.py
MichaelWalker: yes, but I think its a bit trickier as I suspect we interact with this via JS, so I'm going to split it out into its own commit! :)

MichaelWalker: its actually not trickier - we pick it up the current language from the form not from the cookie, so this shouldn't cause any issues. Added it into the PR!

tests/test_middleware.py
```diff
@@ -283,6 +331,45 @@ def test_locale_persist_middleware_deletes_deprecated_cookie(
...
+    settings.LANGUAGE_COOKIE_SECURE = True
+    request = rf.get('/')
+    response = HttpResponse()
+    request.session = client.session
+    instance = middleware.PersistLocaleMiddleware()
+
+    instance.process_response(request, response)
+
+    cookie = response.cookies[settings.LANGUAGE_COOKIE_NAME]
+    assert cookie['httponly'] is True
```
kowalcj0: interesting
https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-CSRF_COOKIE_HTTPONLY

PR #269, opened 2019-08-21
Added full width image component with list of items with title and text

jessgilbert:

 - [x] Changelog entry added.

image
image
image
image


demo/templates/demo/full-width-banners.html
```diff
@@ -77,6 +80,42 @@ <h2 class="heading-large">Hero</h2>
     <pre><code>{% verbatim %} {% case_study background_tint=True background_image=hero_image_domestic heading='Heading text' subtitle='Lorem ipsum dolor sit amet ...'  cta_link='/'  cta_text='Continue' %}{% endverbatim %}</code></pre>
   </div>
 
+  <div class="container">
+    <h3 class="heading-large"> Full width image with a list of items and an optional image or video</h3>
+    <p>The intro is expected to be a mark down field</p>
+    <p>The list of items are expected to be a dictionary with a title and text</p>
+
+    <pre><code>{% verbatim %}title_and_text_dict=[
```
MichaelWalker: `title_and_text_dict` isn't actually a dictionary at the moment, which might cause some confusion.
Could we find another name for them? (we seem to call them `items` in the name of the component, so perhaps that works?)

demo/templates/demo/full-width-banners.html
```diff
@@ -77,6 +80,42 @@ <h2 class="heading-large">Hero</h2>
     <pre><code>{% verbatim %} {% case_study background_tint=True background_image=hero_image_domestic heading='Heading text' subtitle='Lorem ipsum dolor sit amet ...'  cta_link='/'  cta_text='Continue' %}{% endverbatim %}</code></pre>
   </div>
 
+  <div class="container">
+    <h3 class="heading-large"> Full width image with a list of items and an optional image or video</h3>
+    <p>The intro is expected to be a mark down field</p>
```
MichaelWalker: I might be tempted to rename `intro` to `intro_markdown` so that it becomes clear even without reading the docs.

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
```diff
@@ -0,0 +1,42 @@
+{% load add_export_elements_classes from directory_components %}
```
MichaelWalker: I've been trying to come up with a snappier name for this component, but so far I've totally failed! 😂

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
```diff
@@ -0,0 +1,42 @@
...
+
+<section class="full-width-list-with-media " style="background-image: url('{{ background_image }}')">
```
MichaelWalker: The answer to this might just be no, but.....

Is there any room for commonising some behaviour between our 'hero/banner/full-width-image' components?

In my head, they are all a big box with a image background, and a smaller container within it, and the only difference between them is the contents of that container.

Is is easy/possible to extract out a common base similar to what we do with cards, and then each variant just extends the base and overrides what content it contains?

directory_components/export_elements/sass/components/full-width-list-and-media/_base.scss
```diff
@@ -0,0 +1,37 @@
+
+@import "../../partials/colours";
+
+.full-width-list-with-media{
```
MichaelWalker: Probably related to my comment about commonising the components...

If making a base class is tricky, then we can probably at least extract a common mixin for the CSS.
I'd guess most of this CSS is shared with the other similar components?

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
```diff
@@ -0,0 +1,42 @@
...
+    <div class="container padding-top-30 padding-bottom-30">
+        <h1 class="heading-large">{{ title }}</h1>
```
MichaelWalker: Could you make this an `h2`?

The page title will be the `h1`, so this section title should be a `h2`.

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
```diff
@@ -0,0 +1,42 @@
...
+            <div class="column-two-thirds-l">
+                {{ intro|add_export_elements_classes|safe }}
+            </div>
+        </div>
+        <div class="flex-grid reverse-columns-l">
+            <div class="column-half-xl column-full-m">
+                <div class="flex-grid">
+                    {% for item in title_and_text_dict %}
+                        <div class="column-half-l">
+                            <h2 class="bold-medium list-item padding-top-15">{{ item.title }}</h2>
```
MichaelWalker: Likewise could this be bumped down to an `h3`?

MichaelWalker: Given that we are using markdown anyway here, would it make more sense just to let each item be a single markdown field? Or is there a good reason to lock down the form of the item a bit more?

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
```diff
@@ -0,0 +1,42 @@
...
+    <div class="container padding-top-30 padding-bottom-30">
+        <h1 class="heading-large">{{ title }}</h1>
+        <div class="grid-row">
+            <div class="column-two-thirds-l">
+                {{ intro|add_export_elements_classes|safe }}
+            </div>
+        </div>
+        <div class="flex-grid reverse-columns-l">
+            <div class="column-half-xl column-full-m">
+                <div class="flex-grid">
```
MichaelWalker: Conceptually I think we have a list of things here, so the HTML should ideally reflect that.
Could this be a `ul` (and then the divs inside become `li`s)

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
```diff
@@ -0,0 +1,42 @@
...
+                        <div class="column-half-l">
+                            <h2 class="bold-medium list-item padding-top-15">{{ item.title }}</h2>
+                            <span class="font-xsmall">{{ item.text|add_export_elements_classes|safe }}</span>
+                        </div>
+                    {% endfor %}
+                </div>
+            </div>
+            {% if image %}
+            <div class="column-half-xl media">
+                <div class="flex-grid">
```
MichaelWalker: Do we need the grid if there is only a single item?
Is this just to fix the padding etc?

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
```diff
@@ -0,0 +1,42 @@
...
+                    <figure>
+                        <img class="width-full" src="{{ image }}">
+                    </figure>
+                </div>
+            </div>
+            {% elif video %}
+            <div class="column-half-xl media">
+                <div class="flex-grid">
+                    <video controls data-ga-class="hpo-video" class="width-full">
```
MichaelWalker: `data-ga-class` is old and I thought it was dead! 😂 Did you find this kicking about somewhere? If so could you let me know and I'll fix it! 

The video should be tracked in a basic way anyway, so we can probably just get rid of this for now! (`hpo-video` wouldn't be correct in any case)

demo/templates/demo/full-width-banners.html
eadpearce: agree `items` or `items_list` would be better

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
eadpearce: `add_style_classes`? i've also been thinking of renaming this since ye olde export-elements repo is defunct

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
eadpearce: i'd go one step further and suggest this heading level here should be overridable. we do this in some other places already e.g. the message box component

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
jessgilbert: Mike did you mean my very snappy name of `full-width-image-with-list-and-media`? Michal and I also couldn't think of anything better

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
jessgilbert: Oh woops, I stole it from the HPO pages

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
jessgilbert: Yes so its to get display flex and so centering vertically works (margin top and bottom are auto)

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
jessgilbert: Spoke to Bo, he would prefer it was more locked down so there has to be a title which the size can't change and has a small amount of text with a character count. So I think I should keep how it is now as separate fields if that's alright? Will change the h2 to h3 👍

directory_components/templates/directory_components/full-width-image-with-list-and-media.html
jessgilbert: @eadpearce What do you think on this? I haven't done it before so couldn't guess how easy/hard it would be

eadpearce: yeah this would be useful. out of scope of this PR but something I think we should look at soon as we have quite a lot of components now

PR #272, opened 2019-09-02
CI-405: Update International Header.

MichaelWalker:
- Header Items are now set by the app which uses this component.
- Fixed issues with second level navigation.

 - [x] Changelog entry added.

Can still be configured to match the existing header:
image

Or can be configured to fit the new proposed navigation:
image

Mobile View
image

Note that with all the extra data in the new proposed IA I'm the mobile / tablet views get a bit strained very quickly, so we are working to improve that.
It shouldn't cause any regression with the current IA though.


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 28.0.0
+[Full Changelog](https://github.com/uktrade/directory-components/pull/270/files)
+
+### Implemented enhancements
+- Changes to international header to improve support for new sticky second tier navigation.
+
+### Breaking Changes
+- The `InternationalHeader` mixin has been removed, and replaced with the `international_header` templatetag.
```
MichaelWalker: This is potentially controversial, so happy to discuss if it worries anyone!

My logic is that
a) Its more flexible and extendable this way, opening the door for other apps to use the same header pattern.
b) Only 1 app uses the international header now, so it doesn't duplicate any code.
c) The content of the international header now belongs entirely to the international app, which is cleaner and makes it easier to update.

directory_components/templates/directory_components/header_footer/international_header.html
```diff
@@ -22,7 +22,7 @@
       {% include 'directory_components/header_footer/international_header_desktop_nav.html' %}
       {% include 'directory_components/header_footer/mobile_menu_button.html' %}
     </div>
-    {% include 'directory_components/header_footer/international_header_desktop_sub_nav.html' %}
     {% include 'directory_components/header_footer/international_header_mobile_nav.html' %}
   </div>
-</header>
\ No newline at end of file
+</header>
+{% include 'directory_components/header_footer/international_header_desktop_sub_nav.html' %}
```
MichaelWalker: I was slightly sad about this....

Moved outside of the `header` to allow it to be sticky at the top of the page.
Things can only be sticky with respect to their parents, so this sticky thing needs to be a child of the page body which does the scrolling, rather than the header which never needs to scroll.

I think the usability advantages of having it be sticky (UR was very positive about this) outweighs the very slight awkwardness of having it as a separate nav in HTML, but let me know if you have a better idea!

(conceptually its kind of a second nav in any case now, so I think this is fine)

demo/templates/demo/great-international-header-footer.html
```diff
@@ -31,11 +31,55 @@
 
 <section class="padding-top-30">
   <div class="container">
-    <div class="grid-row">
-      <div class="column-two-thirds">
```
eadpearce: please can you keep this in. it makes the body text easier to read on wide screens when it's in a slimmer column

MichaelWalker: sure

demo/views.py
```diff
@@ -37,14 +37,107 @@ def get_context_data(self, *args, **kwargs):
         context['version'] = self.get_version()
         return context
 
+NavNode = namedtuple('NavItem', 'tier_one_item tier_two_items')
+
+def nav_title_to_name(title):
+    return title.lower().replace(' ', '-')
```
eadpearce: `from django.utils.text import slugify` then `slugify(title.lower())`

eadpearce: not actually sure if you even need the `.lower()` i think slugify does that as well

PR #274, opened 2019-09-03
Hero update

eadpearce:

 - [x] Changelog entry added.

Hero

@ desktop
Screen Shot 2019-09-03 at 17 50 45

@ tablet landscape
Screen Shot 2019-09-03 at 17 51 02

@ tablet portrait
Screen Shot 2019-09-03 at 17 51 12

@ mobile
Screen Shot 2019-09-03 at 17 51 25

Hero with CTA (for new great homepage)

@ desktop
Screen Shot 2019-09-05 at 12 14 44

@ tablet landscape
Screen Shot 2019-09-05 at 12 15 05

@ tablet portrait
Screen Shot 2019-09-05 at 12 17 22

@ mobile
Screen Shot 2019-09-05 at 12 17 44


CHANGELOG.md
```diff
@@ -1,5 +1,27 @@
...
+### Breaking Changes
+- Some class names have changed:
+* `case-study` => `great-case-study`
+* `hero` => `great-hero`
+* `hero-with-cta` => `great-hero-with-cta`
+* `hero-title` => `great-hero-title`
+* `hero-text` => `great-hero-text`
+* `heading-hero-generic` => `great-hero-heading`
+* `hero-subheading` => `great-hero-subheading`
+* `hero-description` => `great-hero-description`
```
MichaelWalker: How would you feel about limiting the scope of our CSS in the future so that we don't have these kind of issues?
(ideally changing the styling of a component would be an implementation detail and not a breaking change)

directory_components/export_elements/overrides/govuk_elements/elements/_layout.scss
```diff
@@ -209,6 +234,31 @@ section {
...
+
+  .width-sixth-l {
+    width: percentage(1 / 6);
+  }
+
+  .width-quarter-l {
+    width: 25%;
+  }
+
+  .width-half-l {
```
MichaelWalker: I think these changes should also be considered breaking, as other apps rely on these classes for the layout of elements.

directory_components/export_elements/sass/partials/_typography.scss
```diff
@@ -7,101 +7,79 @@ $quote-font: 'Georgia', serif;
 $line-height: 1.6;
 $hero-line-height: 1.2;
 
-// Great-specific typography mixins
 
-@mixin font-xlarge-normal {
-  font-family: 'Roboto', arial, helvetica, sans-serif;
-  font-size: 24px;
-  font-weight: 400;
+@mixin font-generator($font-family: $brand-font-alternative, $font-size: 19px, $font-size-mobile: 16px, $font-weight: 400, $font-style: normal, $line-height: 1.6) {
```
MichaelWalker: Looks like the default font is the alternative font? Is that right? And if so, should we rename the constant?

richtier:
as this is a breaking change could we take this an opportunity to clean shop a bit: deleting this deprecated template tag https://github.com/uktrade/directory-components/blob/master/directory_components/templatetags/directory_components.py#L112

directory_components/export_elements/sass/partials/_typography.scss
eadpearce: nope. this is intended. brand font is overpass

directory_components/export_elements/overrides/govuk_elements/elements/_layout.scss
eadpearce: these classes didn't exist before so this isn't breaking surely?

CHANGELOG.md
eadpearce: these class name changes only affect components that are written as plain HTML rather than using our template tags so i'm putting this in here as a reference just in case we have an outlier that didn't use the template tag

PR #277, opened 2019-09-06
Add 'key facts' component

eadpearce:

 - [x] Changelog entry added.

@ tablet/desktop
screencapture-components-trade-great-9013-key-facts-2019-09-06-11_57_22

@ mobile
screencapture-components-trade-great-9013-key-facts-2019-09-06-11_27_18


demo/templates/demo/key-facts.html
```diff
@@ -0,0 +1,29 @@
...
+<section class="padding-top-30">
+  <div class="container">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+        <h1 class="heading-xlarge">Key facts</h1>
+        <p>This is a section with max 3 columns each with a heading, text and optional icon.</p>
+      </div>
+    </div>
+  </div>
+</section>
```
MichaelWalker: Would it be easy to add some information about what a 'key_fact' object is?
I think when reading these docs, it would be great to understand what I need to pass in to make the thing work.

directory_components/export_elements/sass/components/key-facts/_base.scss
```diff
@@ -0,0 +1,67 @@
+
+// Key facts component
+// ====================
+
+@import '../../partials/spacings';
+@import '../../partials/mixins';
+
+.great-key-facts {
```
MichaelWalker: deeper question than I really want to answer right now, but...
if we are about to rebrand, is there a better word than `great` to use as our prefix?

MichaelWalker: (this probably isn't something to tackle in this review! :) )

directory_components/templates/directory_components/key_facts.html
```diff
@@ -0,0 +1,8 @@
+<section class="great-key-facts padding-top-60 {% if section_class %}{{ section_class }}{% endif %}">
+  <div class="container wrap-tight">
+    <h2 class="heading-large margin-bottom-30">{{ heading }}</h2>
+    <div class="grid-row">
+      {% include "directory_components/key_facts_item.html" with item=item columns=items|length %}
```
MichaelWalker: wait how does this work? Shouldn't this be `items=items`?

demo/templates/demo/key-facts.html
eadpearce: yep just realised i forgot this

directory_components/templates/directory_components/key_facts.html
MichaelWalker: I think I'd expect this template to contain the looping logic, and the `key_facts_item` to just contain a single item.

directory_components/export_elements/sass/components/key-facts/_base.scss
eadpearce: i thought about this a lot and figured if we do ever rebrand, our styles are also going to change so these classes will likely be rewritten anyway 🤷‍♀

directory_components/templates/directory_components/key_facts.html
eadpearce: LOL i am not actually sure. for some reason it still works

directory_components/templates/directory_components/key_facts.html
```diff
@@ -0,0 +1,8 @@
+<section class="great-key-facts padding-top-60 {% if section_class %}{{ section_class }}{% endif %}">
```
MichaelWalker: is `{% if section_class %}{{ section_class }}{% endif %}` actually just the same as `{{ section_class }}`, or does django blow up if it doesn't exist?

eadpearce: yeah actually i can remove this. was thinking it'd output `None` or something but it doesn't

directory_components/templates/directory_components/key_facts.html
MichaelWalker: I guess the child must just have access to the context of the parent? so `items` was still a thing it could read? Maybe?

eadpearce: yeah no idea. magic

PR #278, opened 2019-09-06
Add details summary component

eadpearce:

 - [x] Changelog entry added.

@ desktop/tablet
closed:
screencapture-components-trade-great-9013-details-2019-09-06-13_50_03

open:
screencapture-components-trade-great-9013-details-2019-09-06-13_49_46

@ mobile:
screencapture-components-trade-great-9013-details-2019-09-06-14_04_34


MichaelWalker:
I thought we had just got rid of all the accordions 😂

eadpearce:
> I thought we had just got rid of all the accordions 😂

LOL talk to michal! though eventually we can phase out ye olde accordion JS and just use details/summary instead. that will be super nice

directory_components/static/directory_components/js/vendor/details-polyfill.js
```diff
@@ -0,0 +1,194 @@
+/*
```
MichaelWalker: Couple of questions

1) Does this also work for Edge? (apparently edge also has issues with details / summary https://caniuse.com/#feat=details

2) What happens without JS in IE. (is this an edge case we need to worry about?)

directory_components/templates/directory_components/base.html
```diff
@@ -44,6 +44,7 @@
         </noscript>
 
         {% block vendor_js %}
+            <script src="{% static 'directory_components/js/vendor/details-polyfill.js' %}"></script>
```
MichaelWalker: Does this mean that on ALL pages we now load a polyfill script that we only use on a very small number of our pages? Are we OK with that as a compromise?

directory_components/static/directory_components/js/vendor/details-polyfill.js
MichaelWalker: actually I think the fallback is just to render everything, always visible, so I think that makes 2) not an issue.

MichaelWalker: https://www.hassellinclusion.com/blog/accessible-accordions-part-2-using-details-summary/

Also seems a bit lukewarm about these from an accessible standpoint, but reading through the article makes me think that its actually fine, as long as we trust that the polyfill does a decent job of making it accessible on IE etc.

eadpearce: 1. Yes
2. It displays as open

MichaelWalker:
This is really cool! Sorry if I've been asking stupid questions - I've read about these tags but never actually used / seen them before!

directory_components/templates/directory_components/base.html
eadpearce: do you want me to put it in a different block?

MichaelWalker: I don't know of an elegant alternative in Django.

Things I've seen done in the past
- Put the script inside the component => but then you need to be careful if there are more than 1 of them on the page
- Just load it on all pages  => but then you have a slight performance hit on every page (although browser cache often limits this)
- Make it so that every page that includes `component-x` also has to add manually `script-x` to the page too => best output, but can be confusing for the dev, (or worse forgotten unless we put in something to prevent that)


I actually don't have an especially strong view on what thing we do - they can all be sensible in the right circumstances. Just wanted to check we'd thought about the implications of what we were doing! 😊

eadpearce: oh right yeah. considering the options yeah it's probably safest/least annoying to put it on all pages. the script isn't particularly big

PR #284, opened 2019-09-11
CMS-1843: Fix language switcher

MichaelWalker:

 - [x] Changelog entry added.

### DIFF
--- a/directory_components/templatetags/directory_components.py
+++ b/directory_components/templatetags/directory_components.py
@@ -227,8 +227,8 @@ def pagination(context, pagination_page, page_param_name='page'):
 HeaderItem = namedtuple('HeaderItem', 'title url is_active')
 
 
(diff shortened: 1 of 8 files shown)


directory_components/templatetags/directory_components.py
```diff
@@ -248,15 +248,16 @@ def international_header(navigation_tree, site_section, site_sub_section):
             ]
 
     return {
+        **context.flatten(),
```
MichaelWalker: The 5 lines in this file are where it all happens! :) 

It seems a little nasty to have to flatten the entire context and force it into an combined dict like this, but I don't have a better suggestion right now - any ideas?

I think the main alternative is explicitly picking out the bits of the context that we need, but 
a) there are quite a few things to pick out
b) its asking for trouble where a dev adds a thing to the 'country_picker' context or something, and forgets to update it here.

PR #289, opened 2019-09-16
GA360 mixin handles request.user and request.sso_user

richtier:
- [x] Changelog entry added.

### DIFF
--- a/makefile
+++ b/makefile
@@ -24,7 +24,7 @@ manage:
 	ENV_FILES='dev' ./manage.py $(ARGUMENTS)
 
 webserver:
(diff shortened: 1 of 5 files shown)


makefile
```diff
@@ -24,7 +24,7 @@ manage:
 	ENV_FILES='dev' ./manage.py $(ARGUMENTS)
 
 webserver:
-	ENV_FILES='dev' python manage.py collectstatic --no-input && ENV_FILES='dev' python manage.py runserver 0.0.0.0:9013 $(ARGUMENTS)
```
richtier: @jessgilbert confirmed the problem that this fixes has gone away. will help fix if it comes back again.

PR #290, opened 2019-09-16
Gave expandable filter button an aria label to fix accessibility

jessgilbert:

 - [x] Changelog entry added.
image

### DIFF
--- a/directory_components/static/directory_components/js/dit.components.searchPageExpandableOptions.js
+++ b/directory_components/static/directory_components/js/dit.components.searchPageExpandableOptions.js
@@ -6,6 +6,13 @@ dit.components.searchPageExpandableOptions = new function(){
         function handleExpandableOptions(event) {
             var element = event.target;
             element.classList.toggle('checked');
+            var aria_label = element.getAttribute('aria-label');
+            if (aria_label.includes("Hide")) {
+                element.setAttribute('aria-label', aria_label.replace("Hide", "Show"));
+            }
+            else if (aria_label.includes("Show")) {
+                element.setAttribute('aria-label', aria_label.replace("Show", "Hide"));
+            }
             var formGroup = document.getElementById(element.getAttribute('aria-controls'));
             formGroup.setAttribute('aria-expanded', !(formGroup.getAttribute('aria-expanded') == 'true'));
         }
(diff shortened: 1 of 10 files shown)


directory_components/templates/directory_components/search-page-expandable-options.html
```diff
@@ -1,6 +1,6 @@
 <div class="expandable-fieldset">
     <label for="toggle_id_{{ field.name }}">{{ label }}</label>
-    <button type="button" id="toggle_id_{{ field.name }}" aria-controls="id_{{ field.name }}-container" class="filter-collapse {{ open|yesno:',checked' }}" alt="{{ alt }}"></button>
+    <button role="button" aria-label="Expandable {{label}} filter" type="button" id="toggle_id_{{ field.name }}" aria-controls="id_{{ field.name }}-container" class="filter-collapse {{ open|yesno:',checked' }}" alt="{{ alt }}"></button>
```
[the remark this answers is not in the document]
jessgilbert: ok 👍

MichaelWalker: Not sure on this, but I _think_ the idea of `aria-label` is to replace the button text in cases where the button doesn't have any text.

So I'd expect the label to follow the same rules as text on a button - ie, they should describe what happens when you click the button, rather than being a description of a button if that makes sense.

So in this case I might change it to something like `{{ open|yesno: 'Show,Hide'}} {{label}} filter options`

MichaelWalker: Not one for today, but in fact I wonder if the absolutely ideal accessible solution is to just always show the options for non-sighted users and to not bother with the toggle at all for them.

ie the point of the toggle is to reduce visual clutter. When we aren't interested in the visuals, it might be best to never hide it. We could make the toggle not visible to screen-readers, and then only visually hide the options.

directory_components/static/directory_components/js/dit.components.searchPageExpandableOptions.js
```diff
@@ -6,6 +6,13 @@ dit.components.searchPageExpandableOptions = new function(){
         function handleExpandableOptions(event) {
             var element = event.target;
             element.classList.toggle('checked');
+            var aria_label = element.getAttribute('aria-label');
```
MichaelWalker: sorry to be a real pain... 
What happens in other languages?

PR #296, opened 2019-09-18
change unordered list bullet colour to #333

jeanettemariemurphy:

 - [x] Changelog entry added.


directory_components/static/directory_components/export_elements/stylesheets/govuk-elements-styles.css
```diff
@@ -1700,16 +1700,44 @@ details {
     color: #006ccc;
     cursor: pointer;
     position: relative;
-    margin-bottom: 0.26316em; }
+    margin-bottom: 0.26316em;
+    padding-left: 25px; }
+    details summary:before {
```
MichaelWalker: It looks like the CSS has changed a lot more than the SCSS.
Is that intended?

PR #298, opened 2019-09-19
Use large image as default

eadpearce:


directory_components/templates/directory_components/hero_with_cta.html
```diff
@@ -1,19 +1,17 @@
...
     <source media="(min-width: 1300px)" srcset="{{ xlarge_image }} 1200w">
     {% endif %}
 
-    {% if large_image %}
     <source media="(min-width: 1025px)" srcset="{{ large_image }} 1200w">
-    {% endif %}
 
     {# default #}
-    <source media="" srcset="{{ medium_image }} 768w">
+    <source media="" srcset="{{ large_image }} 1200w">
```
MichaelWalker: as it stands it looks like we'll never load the medium image. is that deliberate?

directory_components/static/directory_components/export_elements/stylesheets/elements-components.css
```diff
@@ -1494,6 +1494,7 @@ main,
   .great-hero, .great-hero-with-cta {
     display: block; }
     .great-hero > .container, .great-hero-with-cta > .container {
+      height: 100%;
```
MichaelWalker: ah - makes sense! I've seen IE be weird about this before! 😆

demo/templates/demo/full-width-banners.html
```diff
@@ -70,7 +70,7 @@ <h3 class="heading-medium">Responsive image sizes</h3>
       </tr>
     </table>
   </div>
-  <p><code>medium_image</code> must be provided and is the default image size.</p>
+  <p><code>xlarge_image</code> must be provided and is the default image size.</p>
```
MichaelWalker: should this be `large_image`?

MichaelWalker:
Thanks Edie!

Can I check my understanding of this?
- change to the large image default rather than the medium image default as this is a better compromise across all screen sizes
- explicitly set the `height: 100%` to fix IE wierdness 

Does this also fix the weird thing where the title label thing is strangely off to one side?

PR #312, opened 2019-11-14
Improve colour contrast of the error report link and breadcrumbs

richtier:
- [x] Changelog entry added.
 - [x] CSS have been compiled.
 - [x] Change is accessible to the standards we subscribe to.

image

image

image


directory_components/templates/directory_components/base.html
```diff
@@ -109,11 +109,7 @@
             <main id="content" tabindex="-1" class="{% block css_layout_class %}{% endblock css_layout_class %}" role="main">
                 {% block content %}{% endblock %}
                 {% block feedback_reporter %}
-                    <section class="error-reporting" dir="ltr" data-ga-section="feedback">
```
richtier: this is overridden in a few places in order to change the url, so making that easily configurable

directory_components/templates/directory_components/base.html
```diff
@@ -109,11 +109,7 @@
...
-                        <div class="container">
-                            <a id="error-reporting-section-contact-us" href="{{ services_urls.feedback }}" target="_blank">Is there anything wrong with this page?</a>
-                        </div>
-                    </section>
+                    {% include 'directory_components/error-reporting.html' %}
```
richtier: this is overridden in a few places in order to change the url, so making that easily configurable

PR #325, opened 2020-01-28
Feature/mvp ga360 middleware

WillTaylor22:
- [x] Changelog entry added.


directory_components/context_processors.py
```diff
@@ -1,10 +1,24 @@
...
 
 from directory_components import helpers
 
 
+def ga360(request):
+    user = helpers.get_user(request)
+    is_logged_in = helpers.get_is_authenticated(request)
+
+    context = {'ga360': {'site_language': translation.get_language()}}
+    context['ga360']['user_id'] = user.hashed_uuid if user else None
```
richtier: `user` might be an `AnonymouseUser`, so I think this should be something like

`user.hashed_uuid if is_logged_in else None`

directory_components/context_processors.py
```diff
@@ -1,10 +1,24 @@
...
+    context['ga360']['login_status'] = is_logged_in if is_logged_in else False
```
richtier: `ga360']['login_status'] = is_logged_in` should be enough :)

PR #327, opened 2020-01-30
Xot 1290 cookie banner

jeanettemariemurphy:

 - [x] Changelog entry added.
Screenshot 2020-01-30 at 11 14 56
Screenshot 2020-01-30 at 11 14 33


jeanettemariemurphy:
The JS used here is a combination of the one created by digital helpdesk and some elements of the previous js.  I have only customised this a bit as it needs to be aligned with the standards used elsewhere.

PR #329, opened 2020-02-06
Add react cookies modal component

richtier:

 - [x] Changelog entry added.

cookie modal


directory_components/static/directory_components/js/dit.components.cookie-notice.js
```diff
@@ -159,10 +164,9 @@ dit.components.cookieNotice = function() {
...
     createPoliciesCookie: createPoliciesCookie,
-    setPreferencesCookie: setPreferencesCookie
+    setPreferencesCookie: setPreferencesCookie,
+    acceptAllCookies: acceptAllCookies,
+    getPreferencesCookie: getPreferencesCookie,
   };
 }();
 
-window.onload = function() {
-  dit.components.cookieNotice.init('.cookie-notice', '.button-accept', 'cookies');
```
richtier: with this release only the cookie page will need to run this - all other pages will use the new modal (which incidentally uses some of the functions from this file).

PR #336, opened 2020-02-19
preserve querystring in page and refresh page

agiamas:


setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='directory_components',
-    version='35.15.2',
+    version='35.16',
```
richtier: 35.16.0

PR #337, opened 2020-02-26
Feature/allow users without user hashed UUID

WillTaylor22:
- [x] Changelog entry added.

### DIFF
--- a/directory_components/context_processors.py
+++ b/directory_components/context_processors.py
@@ -11,10 +11,8 @@ def ga360(request):
     is_logged_in = helpers.get_is_authenticated(request)
 
     context = {'ga360': {'site_language': translation.get_language()}}
-    if is_logged_in and hasattr(user, 'hashed_uuid'):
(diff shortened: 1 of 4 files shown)


directory_components/context_processors.py
```diff
@@ -11,7 +11,10 @@ def ga360(request):
     is_logged_in = helpers.get_is_authenticated(request)
 
     context = {'ga360': {'site_language': translation.get_language()}}
-    context['ga360']['user_id'] = user.hashed_uuid if is_logged_in else None
```
eadpearce: you could keep this as a one-liner if you wanted `context['ga360']['user_id'] = user.hashed_uuid if is_logged_in and hasattr(user, 'hashed_uuid') else None`

WillTaylor22: Sure, changed back though I tend to avoid ternary operators for anything that isn't super simple as it gets a bit unreadable. This one is probably safe enough though.

PR #351, opened 2021-04-19
Fix tests

webbyfox:


stevejalim (approved):
Makes sense. Would be nice to expand coverage for Magna URLs if possible

PR #356, opened 2021-05-17
Sign out via Magna

rafa-garcia:

 - [ ] Changelog entry added.
 - [ ] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 - [ ] (if updating requirements) Requirements have been compiled.
 - [ ] (if adding CSS) CSS have been compiled.
 - [ ] (if adding HTML) Change is accessible to the standards we subscribe to.
 - [ ] (if adding env vars) Added any new environment variable to vault.
 - [ ] (if adding feature flags) Cleaned up old flags
 - [ ] (if changing the UI) Add a printscreen to the PR

### DIFF
--- a/directory_components/templates/directory_components/header_footer/magna_header.html
+++ b/directory_components/templates/directory_components/header_footer/magna_header.html
@@ -39,12 +39,11 @@
           headers: {
             Accept: 'application/json',
             'Content-Type': 'application/json',
-            'X-CSRFToken': '{{ csrf_token }}',
             'X-Requested-With': 'XMLHttpRequest',
           },
           body: JSON.stringify({})
         }).then(function() {
-          window.location = '/';
+          window.location = '{{ header_footer_urls.magna_home }}';
         });
       });
     }
--- a/directory_components/templates/directory_components/header_footer/magna_header.html
+++ b/directory_components/templates/directory_components/header_footer/magna_header.html
@@ -26,4 +26,27 @@
     {% include 'directory_components/header_footer/magna_header_nav.html' with id='great-header-mobile-nav' device_type='mobile' js_tag=True %}
 
   </div>
+
+  <script>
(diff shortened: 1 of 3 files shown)


directory_components/templates/directory_components/header_footer/magna_header.html
```diff
@@ -26,4 +26,27 @@
...
+          method: 'post',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+            'X-CSRFToken': '{{ csrf_token }}',
+            'X-Requested-With': 'XMLHttpRequest',
+          },
+          body: JSON.stringify({})
+        }).then(function() {
+          window.location = '/';
```
bobmeredith, line 47: I'm wondering where this is going. Isn't this the same as 'magna_home' above if all the webservers are running off the same domain. If they don't, we're dropping to the root of this server's domain.
For consistency, i'd suggest making this {{ header_footer_urls.magna_home }}

directory_components/templates/directory_components/header_footer/magna_header.html
```diff
@@ -26,4 +26,27 @@
...
+
+    if (signOutButton) {
+      signOutButton.addEventListener('click', function(e) {
+        e.preventDefault();
+        fetch(sign_out_endpoint, {
+          method: 'post',
+          headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+            'X-CSRFToken': '{{ csrf_token }}',
```
bobmeredith, line 42: Hmm, not sure about this. Wouldn't this csrf token be generated by webserverx - whereas we are posting to the Magna WS so the tokens wouldn't match?

rafa-garcia: Removing the csrf token from the header as the API endpoint for logging out does not allow cookie-based authentication

bobmeredith (approved):
LGTM

PR #361, opened 2021-09-06
Added international contact us URL

webbyfox:

 - [x] Changelog entry added.

### DIFF
--- a/directory_components/context_processors.py
+++ b/directory_components/context_processors.py
@@ -121,7 +121,8 @@ def invest_header_footer_processor(request):
 def urls_processor(request):
     return {
         'services_urls': {
-            'contact_us': urls.international.CONTACT_US,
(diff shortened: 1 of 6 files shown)


directory_components/context_processors.py
```diff
@@ -121,7 +121,7 @@ def invest_header_footer_processor(request):
 def urls_processor(request):
     return {
         'services_urls': {
-            'contact_us': urls.domestic.CONTACT_US,
+            'contact_us': urls.international.CONTACT_US,
```
stevejalim, line 124: Do we definitely need to do this? I thought we could maybe just override in the international footer only? Happy to chat tomorrow if needed

webbyfox: I don't think there is a way we can simply override via env variable.  Also, domestic is no more needed hence changed it to international.

stevejalim (approved):

PR #374, opened 2023-05-22
bump django to 4.1.9

davidu1975:

 - [x] Changelog entry added.
 - [x] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 - [x] (if updating requirements) Requirements have been compiled.


setup.py
```diff
@@ -13,26 +13,49 @@
     long_description_content_type='text/markdown',
     include_package_data=True,
     install_requires=[
-        'django>=1.11,<4.0.0',
+        'django>=1.11,<=4.1.9',
```
mforner13, line 16: Let's do 4.2 here so we don't have to update again when we move to 4.2 in the future :)

davidu1975: done

mforner13 (approved):

PR #383, opened 2024-01-30
KLS-1795 - add block for inline feedback inside body tags

timothyPatterson:
This PR adds a block to contain inline feedback. The addition is necessary to enable content to be added within body tags that would not be considered part of _main_ content.


lewis-coulson-dit (approved):

hal274 (approved):
LGTM

